PR #3, opened 2020-01-22
Make a CI testing process

sjmiller609:
... and also patch for compatibility with up to k8s version 1.17

After merging, enable circle CI for this repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ workflows:
     jobs:
       - 'airflow-on-k8s-1.14'
       - 'airflow-on-k8s-1.15'
-      # - 'airflow-on-k8s-1.16'
-      # - 'airflow-on-k8s-1.17'
+      - 'airflow-on-k8s-1.16'
+      - 'airflow-on-k8s-1.17'
 
 jobs:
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,56 @@
+version: 2
+
+workflows:
+  version: 2
+  install-airflow-chart:
+    jobs:
+      - 'airflow-on-k8s-14'
+      - 'airflow-on-k8s-15'
+      # - 'airflow-on-k8s-16'
+      # - 'airflow-on-k8s-17'
```
sjmiller609: Future PRs will come to enable support for newest kube versions. These are not working at the moment.

templates/statsd/statsd-deployment.yaml
```diff
@@ -2,7 +2,7 @@
 ## Airflow StatsD Deployment
 #################################
 kind: Deployment
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
```
sjmiller609: this is the only change required for compatibility in 1.16 and 1.17. still compatible back to 1.14

PR #5, opened 2020-01-23
Identify ssl mode in conn string

dimberman:
For using conn string in KEDA we need to identify the sslmode


schnie:
Nice, looks good! Should we also add the same config to the result backend connection for celery? https://github.com/astronomer/airflow-chart/blob/master/templates/secrets/result-backend-connection-secret.yaml#L19

dimberman:
@schnie added resultBackendConn

PR #6, opened 2020-01-23
Add a release process to CI

sjmiller609:
This will require a secret added to the pipeline such that gsutil is authorized to upload to gs://helm.astronomer.io

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+set -e
+
+if [[ -z "${HELM_CHART_PATH}" ]]; then
+  echo "ERROR: must provide HELM_CHART_PATH pointing to airflow-*.tgz, will not release without an artifact."
+  exit 1
+fi
+
+# Common bash line that sets DIR to the directory of this script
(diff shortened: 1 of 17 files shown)


bin/release
```diff
@@ -0,0 +1,68 @@
...
+echo "========================="
+echo ""
+
+echo "Releasing chart..."
+set -xe
+# Set up the index.html file
+wget https://helm.astronomer.io/index.yaml -O /tmp/index.yaml.current
+helm repo index ./airflow --url https://helm.astronomer.io --merge /tmp/index.yaml.current
+gsutil cp -a public-read $HELM_CHART_PATH gs://helm.astronomer.io
+gsutil cp -a public-read ./airflow/index.yaml gs://helm.astronomer.io
```
sjmiller609: I expect we will manually pull the index.yaml and delete the old subcharts eventually.

sjmiller609: this block is why we need the environment variable to authenticate gsutil

sjmiller609:
I'm going to want to add github release also... sec

PR #9, opened 2020-01-28
Add support of path based airflow ingress

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/459

### DIFF
--- a/bin/run-ci
+++ b/bin/run-ci
@@ -1,6 +1,26 @@
 #!/bin/bash
 set -e
 
+function get_debugging_info {
+  echo "Failed to deploy Airflow!"
+  echo "Printing description and logs where containers in pod are not ready..."
+  for pod in $(kubectl get pods -n airflow | grep -v NAME | grep -v 1/1 | grep -v 2/2 | grep -v Completed | awk '{ print $1 }'); do
+    echo "======================="
+    set -x
+    kubectl describe pod -n airflow $pod
+    kubectl logs -n airflow $pod | tail -n 30
+    set +x
+    echo "======================="
+  done
+  kubectl get pods --all-namespaces
+  echo "======================="
+  kubectl get ds --all-namespaces
+  echo "======================="
+  kubectl get sts --all-namespaces
+  echo "======================="
+  kubectl get deployments --all-namespaces
+}
+
 # Common bash line that sets DIR to the directory of this script
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 REPO_DIR=$DIR/../
@@ -18,15 +38,22 @@ WATCH_PID=$!
 
 START=$(date +%s)
 
+set +e
 set -x
+
 helm install \
   -n airflow \
   --wait \
   $HELM_CHART_PATH
 
-HELM_CODE=$?
+if [ ! $? -eq 0 ]; then
+  echo "Helm chart failed to install"
+  get_debugging_info
+  exit 1
+fi
+
+HELM_CODE=0
 set +x
-kill $WATCH_PID
 
 # Wait for all pods to become ready
 RC=0
@@ -52,24 +79,13 @@ while [ $RC -eq 0 ]; do
   fi
 done
 
+kill $WATCH_PID
+
 # Output debugging information if the Helm install failed
 if [ $HELM_CODE -eq 0 ]; then
   echo "Airflow deployed!"
 else
   echo "Failed to deploy Airflow!"
-  echo "Printing description and logs where containers in pod are not ready..."
-  for pod in $(kubectl get pods | grep -v NAME | grep -v 1/1 | grep -v 2/2 | awk '{ print $1 }'); do
-    echo "======================="
-    set -x
-    kubectl describe pod $pod
-    kubectl logs $pod | tail -n 30
-    set +x
-    echo "======================="
-  done
+  get_debugging_info
+  exit 1
 fi
-
-set -x
-kubectl get pods --all-namespaces
-helm list
-
-exit $HELM_CODE
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -78,12 +78,16 @@
   {{ end }}
 {{- end }}
 
+{{ define "airflow_host" -}}
+  {{ printf "deployments.%s" .Values.ingress.baseDomain }}
+{{- end }}
+
 {{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+  {{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
 {{- end }}
 
 {{ define "flower_url" -}}
-{{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }}
+  {{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }}
 {{- end }}
 
 # This helper will change when customers deploy a new image.
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -78,16 +78,16 @@
   {{ end }}
 {{- end }}
 
-{{ define "airflow_host" -}}
-  {{ printf "deployments.%s" .Values.ingress.baseDomain }}
+{{ define "deployments_subdomain" -}}
+{{ printf "deployments.%s" .Values.ingress.baseDomain }}
 {{- end }}
 
-{{ define "airflow_url" -}}
-  {{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+{{ define "airflow_subdomain" -}}
+{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
 {{- end }}
 
-{{ define "flower_url" -}}
-  {{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }}
+{{ define "flower_subdomain" -}}
+{{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }}
 {{- end }}
 
 # This helper will change when customers deploy a new image.
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -27,7 +27,7 @@ data:
 
     [webserver]
 {{- if .Values.ingress.enabled }}
-    base_url = {{ printf "https://%s/%s/airflow" (include "airflow_host" .) .Release.Name }}
+    base_url = {{ printf "https://%s/%s/airflow" (include "deployments_subdomain" .) .Release.Name }}
 {{- end }}
     enable_proxy_fix = True
     expose_config = True
@@ -97,7 +97,7 @@ data:
     [astronomer]
 {{- if .Values.webserver.jwtSigningCertificateSecretName }}
     jwt_signing_cert = {{ template "airflow_webserver_jwt_cert_path" . }}
-    jwt_audience = {{ printf "%s/%s" ( include "airflow_host" . ) .Release.Name }}
+    jwt_audience = {{ printf "%s/%s" ( include "deployments_subdomain" . ) .Release.Name }}
 {{- end }}
 
   webserver_config.py: |
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -26,7 +26,9 @@ data:
 {{- end }}
 
     [webserver]
+{{- if .Values.ingress.enabled }}
     base_url = {{ printf "https://%s/%s/airflow" (include "airflow_host" .) .Release.Name }}
+{{- end }}
     enable_proxy_fix = True
     expose_config = True
     rbac = True
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -26,7 +26,8 @@ data:
 {{- end }}
 
     [webserver]
-    base_url = {{ printf "https://%s" (include "airflow_url" .) }}
+    base_url = {{ printf "https://%s/%s/airflow" (include "airflow_host" .) .Release.Name }}
+    enable_proxy_fix = True
     expose_config = True
     rbac = True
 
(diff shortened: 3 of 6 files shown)


templates/_helpers.yaml
```diff
@@ -79,11 +79,11 @@
 {{- end }}
 
 {{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+app.local.astronomer-development.com
```
ashb: This change is wrong for non-dev I think

templates/webserver/webserver-deployment.yaml
```diff
@@ -88,22 +88,22 @@ spec:
...
-            httpGet:
-              path: /health
-              port: {{ .Values.ports.airflowUI }}
-            initialDelaySeconds: {{ .Values.webserver.readinessProbe.initialDelaySeconds | default 15 }}
-            timeoutSeconds: {{ .Values.webserver.readinessProbe.timeoutSeconds | default 30 }}
-            failureThreshold: {{ .Values.webserver.readinessProbe.failureThreshold | default 20 }}
-            periodSeconds: {{ .Values.webserver.readinessProbe.periodSeconds | default 5 }}
+#          livenessProbe:
+#            httpGet:
+#              path: /health
```
ashb: These should still be enabled

templates/ingress.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
     kubernetes.io/tls-acme: {{ eq .Values.ingress.acme true | quote }}
     nginx.ingress.kubernetes.io/custom-http-errors: "403,404,502,503"
   {{- if .Values.ingress.auth.enabled }}
-    nginx.ingress.kubernetes.io/auth-url: https://houston.{{ .Values.ingress.baseDomain }}/v1/authorization
+    nginx.ingress.kubernetes.io/auth-url: http://astronomer-houston.astronomer.svc.cluster.local:8871/v1/authorization
```
ashb: This too

templates/_helpers.yaml
schnie: These could both be `printf "app.%s" .Values.global.baseDomain`

templates/webserver/webserver-deployment.yaml
schnie: Agreed, I think our `base_url` needs to be accounted for here.

templates/ingress.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
     kubernetes.io/tls-acme: {{ eq .Values.ingress.acme true | quote }}
     nginx.ingress.kubernetes.io/custom-http-errors: "403,404,502,503"
   {{- if .Values.ingress.auth.enabled }}
-    nginx.ingress.kubernetes.io/auth-url: https://houston.{{ .Values.ingress.baseDomain }}/v1/authorization
+    nginx.ingress.kubernetes.io/auth-url: http://astronomer-houston.{{ .Values.platform.namespace }}.svc.cluster.local:8871/v1/authorization
```
andriisoldatenko: @ashb @schnie i have some troubles with this line.

ashb: I think outside of your local testing this line should need changing at all.

templates/configmap.yaml
```diff
@@ -26,7 +26,8 @@ data:
 {{- end }}
 
     [webserver]
-    base_url = {{ printf "https://%s" (include "airflow_url" .) }}
+    base_url = {{ printf "https://%s/deployments/%s/airflow" (include "airflow_url" .) .Release.Name }}
```
ashb: This should be in the airflow_url macro, not here.

templates/_helpers.yaml
```diff
@@ -79,11 +79,11 @@
 {{- end }}
 
 {{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+  {{ printf "app.%s" .Values.ingress.baseDomain }}
```
ashb: I don't think airflow should live under the same subdomain as "app"/oribit -- that gives it access to the Orbit auth cookie etc and is something we should avoid.

What do you think about a subdomain dedicated to airflow/flowers/deployments?

templates/ingress.yaml
schnie: Yep, agreed

templates/_helpers.yaml
andriisoldatenko: @schnie do you have requirements for sub-domain name?
```
https://airflow.{{ .Values.ingress.baseDomain }}/deployments/release-name-123/airflow ?
```

templates/ingress.yaml
andriisoldatenko: fixed.

templates/_helpers.yaml
```diff
@@ -79,11 +79,12 @@
...
 
 {{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+  # TODO: @andriisoldatenko make sure we are all confirm airflow subdomain
+  {{ printf "airflow.%s/deployments/%s/airflow" .Values.ingress.baseDomain .Release.Name }}
 {{- end }}
 
 {{ define "flower_url" -}}
-{{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }}
+  {{ printf "app.%s" .Values.ingress.baseDomain }}
```
ashb: This one too -- where is this used?

templates/_helpers.yaml
andriisoldatenko:
```
https://deployments.{{ .Values.ingress.baseDomain }}/d/release-name-123/airflow
https://deployments.{{ .Values.ingress.baseDomain }}/d/release-name-123/flower
```
confirmed.

templates/_helpers.yaml
andriisoldatenko: @ashb 
templates/ingress.yaml:54

andriisoldatenko:
@ashb @schnie PTAL

templates/_helpers.yaml
```diff
@@ -78,12 +78,16 @@
   {{ end }}
 {{- end }}
 
+{{ define "deployments_host" -}}
+  {{ printf "deployments.%s" .Values.ingress.baseDomain }}
+{{- end }}
+
 {{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+  {{ printf "deployments.%s/%s/airflow" .Values.ingress.baseDomain .Release.Name }}
```
ashb: Can you call one macro from another? If so using `deployments_host` here would be good.

templates/configmap.yaml
```diff
@@ -94,7 +95,7 @@ data:
     [astronomer]
 {{- if .Values.webserver.jwtSigningCertificateSecretName }}
     jwt_signing_cert = {{ template "airflow_webserver_jwt_cert_path" . }}
-    jwt_audience = {{ include "airflow_url" . }}
+    jwt_audience = {{ include "deployments_host" . }}
```
ashb: This needs to change - see https://github.com/astronomer/houston-api/pull/211#pullrequestreview-352285616

templates/configmap.yaml
```diff
@@ -94,7 +95,7 @@ data:
     [astronomer]
 {{- if .Values.webserver.jwtSigningCertificateSecretName }}
     jwt_signing_cert = {{ template "airflow_webserver_jwt_cert_path" . }}
-    jwt_audience = {{ include "airflow_url" . }}
+    jwt_audience = {{ printf "%s/%s" ( include "deployments_host" . ) .Release.Name }}
```
andriisoldatenko: @ashb ☝️

andriisoldatenko:
@ashb @schnie PTAL!

templates/ingress.yaml
```diff
@@ -23,6 +23,14 @@ metadata:
     nginx.ingress.kubernetes.io/auth-url: https://houston.{{ .Values.ingress.baseDomain }}/v1/authorization
     nginx.ingress.kubernetes.io/auth-signin: https://app.{{ .Values.ingress.baseDomain }}/login
     nginx.ingress.kubernetes.io/auth-response-headers: authorization, username, email
+    nginx.ingress.kubernetes.io/proxy-cookie-path: / /{{ .Release.Name }}/
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      if ($host = '{{- include "old_airflow_url" . }}' ) {
+        rewrite ^ https://deployments.{{ .Values.ingress.baseDomain }}/{{ .Release.Name }}/airflow permanent;
```
schnie: @andriisoldatenko could these snippets be shortened to use your new `deployments_host` helper?

templates/_helpers.yaml
```diff
@@ -78,12 +78,16 @@
...
 
-{{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+{{ define "deployments_host" -}}
+  {{ printf "deployments.%s" .Values.ingress.baseDomain }}
 {{- end }}
 
-{{ define "flower_url" -}}
-{{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }}
+{{ define "old_airflow_url" -}}
```
schnie: @andriisoldatenko what do you think about using `airflow_host`, `airflow_url`, `flower_host` and `flower_url` for these helper names? Instead of using `old` here?

Looks like we do a combination of creating urls/paths inside our helpers, as well as directly inline in `ingress.yaml`. We could re-use the helpers there, as well as in the `configmap.yaml`.

templates/ingress.yaml
andriisoldatenko: fixed.

templates/_helpers.yaml
andriisoldatenko: @schnie @ashb looks like last point before approve?
not sure that i get Greg's point.

andriisoldatenko: @schnie after chatting with @ashb i think i understand your point:
refactor `<.Release.Name>/airflow` to `(include airflow-path .)`?
IMO ver1 `{{ .Release.Name>/airflow }}` is more readable than,
ver2: `{{ printf "%s/health" (include airflow-path .) }}`

andriisoldatenko:
waiting for this PR https://github.com/astronomer/ap-airflow/pull/36 to see green CI.

sjmiller609:
@andriisoldatenko my PR only adds the test - since the test is passing, I think that means the existing master branch has the correct version (so you are not blocked on my PR). Let me check what has been released dockerhub.

sjmiller609:
I think the master branch is released to dockerhub.

sjmiller609:
the issue is that we are using "latest" by default:
https://github.com/astronomer/airflow-chart/blob/270e0b2c59624dab3faadba127da11e58a6de14f/values.yaml#L67-L70
but "latest" is not a tag that's updated by our CI https://hub.docker.com/r/astronomerinc/ap-airflow/tags?page=1&name=latest

schnie:
There's some more logic defined in the `airflow_image` helper regarding the tag.

sjmiller609:
Ok I see:
https://github.com/astronomer/airflow-chart/blob/270e0b2c59624dab3faadba127da11e58a6de14f/templates/webserver/webserver-deployment.yaml#L57
https://github.com/astronomer/airflow-chart/blob/270e0b2c59624dab3faadba127da11e58a6de14f/templates/_helpers.yaml#L95-L97
https://github.com/astronomer/airflow-chart/blob/270e0b2c59624dab3faadba127da11e58a6de14f/values.yaml#L59-L63

I am mistaken about 'latest'.

However the image contains the wrong thing:

```
$ docker run -it --rm astronomerinc/ap-airflow:1.10.7-alpine3.10 pip freeze | grep redis
redis==3.4.0
```

sjmiller609:
So I think we just need to re-build and re-publish the images, which can be accomplished by merging my PR on the airflow-chart repo to master.

andriisoldatenko:
@sjmiller609 yes please.

sjmiller609:
@ashb @andriisoldatenko so I believe since nothing changed in the master branch, and the currently published version includes redis 3.4.0 and now it does not (with identical Dockerfile) that the build is not reproducible because the module versions are unconstrained. I think this is already a known thing about airflow installation, right?

I made this issue to attempt to prevent this kind of easily detectable issue in the future (crashlooping pods / helm chart install not working): https://github.com/astronomer/issues/issues/758

bin/run-ci
```diff
@@ -1,6 +1,26 @@
...
+  done
+  kubectl get pods --all-namespaces
+  echo "======================="
+  kubectl get ds --all-namespaces
+  echo "======================="
+  kubectl get sts --all-namespaces
+  echo "======================="
+  kubectl get deployments --all-namespaces
+}
+
```
sjmiller609: I added this block to help with debugging CI failures. this is redundant with bin/run-ci in the Astronomer chart and they will be de-duplicated soon enough.

templates/_helpers.yaml
```diff
@@ -78,12 +78,16 @@
   {{ end }}
 {{- end }}
 
+{{ define "airflow_host" -}}
+  {{ printf "deployments.%s" .Values.ingress.baseDomain }}
+{{- end }}
+
 {{ define "airflow_url" -}}
-{{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
+  {{ .Release.Name }}-airflow.{{ .Values.ingress.baseDomain }}
```
schnie: @andriisoldatenko can we un-indent the helpers in here? I think "standard" practice is no indentation on these. For example: https://github.com/helm/charts/blob/master/stable/prometheus/templates/_helpers.tpl

schnie: @andriisoldatenko thinking more about these helper names, sorry!

What do you think about these? Just trying to come up with something that is consistent and makes sense.
`airflow_host` = `deployments_subdomain`
`airflow_url` = `airflow_subdomain`
`flower_url` to `flower_subdomain`

andriisoldatenko: @schnie ptal

saikatharryc:
any idea why the airflow dashboard URI badly formatted?

```
 https://deployments.{{ .Values.ingress.baseDomain }}/{{ .Release.Name }}/airflow
```

cant it be something as simple as this: ` https://airflow.{{ .Values.ingress.baseDomain }}/`

schnie:
@saikatharryc this chart is used by the Astronomer Platform and needs to use these URLs.

The good news is that we recently contributed this same chart to the official Apache Airflow project, which lives here - https://github.com/apache/airflow/tree/master/chart.

Any improvements around ingress would be much appreciated!

PR #12, opened 2020-01-30
Specify defaults for resources

sjmiller609:
These are the changes that Tony did.

**Checklist**


schnie:
I didn't realize these were empty by default. I think that makes more sense for open source. Lets override at cloud level.

sjmiller609:
https://github.com/astronomer/terraform-google-astronomer-cloud/pull/119

PR #13, opened 2020-01-30
Add option to override airflow_local_settings.py

sjmiller609:
Some users may need to override the Airflow pod mutation hook.

Example of this change at work:
```
e4384cda2b17:  ~/shared/airflow-chart 
_ helm template . -f ./config.yaml | grep -A9 'airflow_local_settings.py:'
  airflow_local_settings.py: |

    from airflow.contrib.kubernetes.pod import Pod
    from airflow.configuration import conf

    def pod_mutation_hook(pod: Pod):
      pass


---

e4384cda2b17:  ~/shared/airflow-chart 
_ cat config.yaml
scheduler:
  airflowLocalSettingsOverride: |
    from airflow.contrib.kubernetes.pod import Pod
    from airflow.configuration import conf

    def pod_mutation_hook(pod: Pod):
      pass

e4384cda2b17:  ~/shared/airflow-chart 
_ helm template . | grep -A25 'airflow_local_settings.py:'
  airflow_local_settings.py: |
    from airflow.contrib.kubernetes.pod import Pod
    from airflow.configuration import conf


    def pod_mutation_hook(pod: Pod):

        extra_labels = {
            "kubernetes-executor": "False",
            "kubernetes-pod-operator": "False"
        }

        if 'airflow-worker' in pod.labels.keys() or \
                conf.get('core', 'EXECUTOR') == "KubernetesExecutor":
            extra_labels["kubernetes-executor"] = "True"
        else:
            extra_labels["kubernetes-pod-operator"] = "True"

        pod.labels.update(extra_labels)
        pod.tolerations += []
        pod.affinity.update({})

---
# Source: airflow/templates/scheduler/scheduler-serviceaccount.yaml
## Airflow Scheduler ServiceAccount

e4384cda2b17:  ~/shared/airflow-chart 
```
**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -187,6 +187,12 @@ scheduler:
   #   cpu: 100m
   #   memory: 128Mi
 
+  # This setting can overwrite
+  # podMutation settings - be sure
+  # to reference the default if you
+  # are making use of those settings.
+  airflowLocalSettingsOverride: ~
+
 # Airflow webserver settings
 webserver:
   livenessProbe:
--- a/values.yaml
+++ b/values.yaml
@@ -191,7 +191,7 @@ scheduler:
   # podMutation settings - be sure
   # to reference the default if you
   # are making use of those settings.
-  airflowLocalSettingsOverride: ~
+  airflowLocalSettings: ~
 
 # Airflow webserver settings
 webserver:
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v1 is Helm 2
 apiVersion: v1
 name: airflow
-version: 0.11.0
+version: 0.11.1
 description: Helm chart to deploy the Astronomer Platform Airflow module
 icon: https://www.astronomer.io/static/airflowNewA.png
 keywords:
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -119,9 +119,9 @@ data:
     SECURITY_MANAGER_CLASS = AirflowAstroSecurityManager
 {{- end }}
 
-{{- if .Values.scheduler.airflowLocalSettingsOverride }}
+{{- if .Values.scheduler.airflowLocalSettings }}
   airflow_local_settings.py: |
-    {{ .Values.scheduler.airflowLocalSettingsOverride | nindent 4 }}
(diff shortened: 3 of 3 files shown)


values.yaml
```diff
@@ -187,6 +187,12 @@ scheduler:
   #   cpu: 100m
   #   memory: 128Mi
 
+  # This setting can overwrite
+  # podMutation settings - be sure
+  # to reference the default if you
+  # are making use of those settings.
+  airflowLocalSettingsOverride: ~
```
schnie: @sjmiller609 what do you think about shortening to `airflowLocalSettings` and dropping `Override` for consistency. `Override` can become redundant so we've opted to just name things precisely what they are.

sjmiller609: yes that's better.

sjmiller609:
> Looks good. What triggered this initial use-case? I know we talked yesterday, but I think my brain had already quit for the day.

k8s executor and KPO pods are getting their command (kube) / entrypoint (docker) overwritten by the scheduler. so 'airflow' is getting called directly instead of waiting for istio-proxy. We can control this behavior by modifying the Airflow pod mutation hook.

sjmiller609:
@schnie I gave madeline ETA monday for the k8s executor and kpo fix

andrewhharmon:
How’s this one looking? After it’s merged, is there still a wait for a patch release, or can it be pushed out? This one’ s looming kind of heavy for a few customers.

schnie:
@sjmiller609 can you paste an example of what we'd need this setting to look like in our cloud?

schnie:
@sjmiller609 do we need to make a PR to https://github.com/astronomer/terraform-google-astronomer-cloud to use this new feature? Would be helpful to see that as well.

sjmiller609:
> @sjmiller609 do we need to make a PR to https://github.com/astronomer/terraform-google-astronomer-cloud to use this new feature? Would be helpful to see that as well.

Yes it will require a change. I did not develop the modified mutation hook yet, so there is no code to see at this point. I was waiting to do that after this change is available.

sjmiller609:
I believe that our use-case is not the only reason someone could want this feature. That is outside the scope of this change.

PR #15, opened 2020-02-05
Fix installation path of CI tool

sjmiller609:
**Checklist**

- [ ]  Chart.yaml version updated


sjmiller609:
this worked: https://github.com/sjmiller609/test-circle-temporary-repo/blob/master/.circleci/config.yml

PR #16, opened 2020-02-05
Fix python version error

sjmiller609:
The error was
```
pyenv: python2: command not found

The `python2' command exists in these Python versions:
  2.7.12
```
~~I did not test this will fix it. I'm not sure how to do so...~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,8 +107,7 @@ jobs:
           name: Release the Airflow chart
           command: |
             set -e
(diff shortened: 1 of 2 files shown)


sjmiller609:
my guess for why is failed was the gcloud was installed using python2 by default in the ubuntu machine CI environment.

sjmiller609:
I want to test it more

sjmiller609:
so yeah, I think that circle CI basically had gcloud installed using python2's pip. So we had the gcloud command at the beginning and that's caused the issue I think. I did reproduce this in my own Circle.

PR #22, opened 2020-02-05
Disable the scheduler PodDisruptionBudget

schnie:
This PR makes the PDB on the scheduler and pgbouncer optional and disabled by default. By disabling in our cloud we may be able to avoid needing to do the manual compaction.

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -175,9 +175,13 @@ workers:
 
 # Airflow scheduler settings
 scheduler:
-  # Ensure we have one replica always running
   podDisruptionBudget:
-    maxUnavailable: 1
+    enabled: false
+
+    # PDB configuration
+    config:
+      # Ensure we have one replica always running
+      maxUnavailable: 1
 
   resources: {}
   #  limits:
--- a/values.yaml
+++ b/values.yaml
@@ -175,12 +175,12 @@ workers:
 
 # Airflow scheduler settings
 scheduler:
+  # Scheduler pod disruption budget
   podDisruptionBudget:
     enabled: false
 
     # PDB configuration
     config:
-      # Ensure we have one replica always running
       maxUnavailable: 1
 
   resources: {}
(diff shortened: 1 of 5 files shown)


schnie:
@verdverm what do you think about this?

schnie:
@ashb I think one of the reasons we have to manually compact our node pools every once in a while, is that we have these pods out there with PDBs, so the autoscaler won't kill those pods to scale down a node. 

So we end up with underutilized nodes when people are deleting deployments one way or the other.

sjmiller609:
@schnie we still have kubecost running. IIRC it includes tips on why nodes are not getting compacted.

values.yaml
```diff
@@ -175,9 +175,13 @@ workers:
...
 # Airflow scheduler settings
 scheduler:
-  # Ensure we have one replica always running
   podDisruptionBudget:
-    maxUnavailable: 1
+    enabled: false
+
+    # PDB configuration
+    config:
+      # Ensure we have one replica always running
```
verdverm: This comment is likely misleading

verdverm:
This should enable the pods to be drained / evicted, but can cause disruption if the service (graph) is not resilient to forceful exit / cancellation of open sockets. A k8s Deployment rollingUpdate values also interact with this process. StatefulSets are more complicated because they are indexed and want to recover their PVC.

For `pgbouncer`, we could default to no PDB and add a `maxSurge: 1` here: https://github.com/astronomer/airflow-chart/blob/master/templates/pgbouncer/pgbouncer-deployment.yaml#L21  This should create a new instance before removing the old one, and ought to be ok since pgbouncer is stateless, and the application talks to it through a k8s service.

For `scheduler` we do need the PDB when stateful, can probably use the pgbouncer method if not.

---

Some more useful links:

- https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
- https://kubernetes.io/docs/tasks/run-application/configure-pdb/#think-about-how-your-application-reacts-to-disruptions

schnie:
Good call @verdverm. Added the `maxSurge: 1` to pgbouncer `Deployment`.

schnie:
cc: @vishwas-astro @vikramkoka @ryw this is something that could potentially help us scale nodes down more aggressively without tweaking the autoscaler just yet.

PR #25, opened 2020-02-07
Adds safe-to-evict annotations to pods that use emptyDir

schnie:
This PR adds support to optionally flag some pods as safe-to-evict, in the event that kubernetes needs to scale a node down (or some other cases). This initial commit flags the scheduler, workers and redis as safe-to-evict, but allows us to override from cloud if needed.

In addition to https://github.com/astronomer/issues/issues/763 (Removal of PodDisruptionBudgets), this increases our aggressiveness in killing pods to scale down.

These are all pods that could potentially make use of `emptyDir` volumes. These are volumes that exist on a particular node, until that pod is finally relocated. These volumes are cleaned up after a relocation, and we currently cap all pods to a few Gi of this ephemeral storage.

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node


schnie:
cc: @vikramkoka @vishwas-astro @ryw

schnie:
@sjmiller609 the `safeToEvict` flags are only nested one layer under keys that have existed forever, so these should not introduce an upgrade issue.

templates/redis/redis-statefulset.yaml
```diff
@@ -31,6 +31,10 @@ spec:
 {{- with .Values.labels }}
 {{ toYaml . | indent 8 }}
 {{- end }}
+      annotations:
+        {{- if .Values.redis.safeToEvict }}
```
schnie: As we upgrade charts, this will be `null`, so evaluate to `false` and not show, unless we explicitly override from Houston's `ConfigMap`.

PR #26, opened 2020-02-07
Fix to match new naming convention for KEDA 1.2

dimberman:
**Checklist**


sjmiller609:
I had to enable Circle for forks. now it's on, but it will not provide secrets to the build.

PR #32, opened 2020-02-15
Add support for Azure DB PostgreSQL - airflow chart

ianstanton:
fix issue astronomer/issues#102
- Adds username in format `user@dbserver` and password to `pgbouncer_config` if `azureDbServer` value exists for `metadataConnection` or `resultBackendConnection`
- `azureDbServer` value will be added by `houston-api` if it exists in the username https://github.com/astronomer/houston-api/pull/233
- Required for use with Azure DB for PostgreSQL

### DIFF
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -154,8 +154,8 @@
 
 {{ define "pgbouncer_config" }}
 [databases]
-{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }}
-{{ .Release.Name }}-result-backend = host={{ .Values.data.resultBackendConnection.host }} dbname={{ .Values.data.resultBackendConnection.db }} port={{ .Values.data.resultBackendConnection.port }} pool_size={{ .Values.pgbouncer.resultBackendPoolSize }}
+{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }} {{ if .Values.data.metadataConnection.azureDbServer }}user={{ .Values.data.metadataConnection.user }}{{ .Values.data.metadataConnection.azureDbServer }} password={{ .Values.data.metadataConnection.pass }}{{ end }}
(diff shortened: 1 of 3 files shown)


templates/_helpers.yaml
```diff
@@ -154,8 +154,8 @@
 
 {{ define "pgbouncer_config" }}
 [databases]
-{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }}
-{{ .Release.Name }}-result-backend = host={{ .Values.data.resultBackendConnection.host }} dbname={{ .Values.data.resultBackendConnection.db }} port={{ .Values.data.resultBackendConnection.port }} pool_size={{ .Values.pgbouncer.resultBackendPoolSize }}
+{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }} {{ if .Values.data.metadataConnection.azureDbServer }}user={{ .Values.data.metadataConnection.user }}{{ .Values.data.metadataConnection.azureDbServer }} password={{ .Values.data.metadataConnection.pass }}{{ end }}
```
schnie: @ianstanton is there a way we could standardize to a format that works for typical behavior and azure behavior, rather than adding `user` and `password` _only_ for azure? Could we do that for both?

schnie: @ianstanton I'm hoping that this would also let us drop `azureDbServer` out of houston, keeping the logic generic and consolidated.

ianstanton:
cc @vishwas-astro

ianstanton:
@schnie I've updated this to always pass `user` and `password` to the `databases` section of `pgbouncer.ini`. 

The `users.txt` file is still required in this case. If we want to drop the `users.txt` file, we'll need to set `auth_type = any`

- auth_type
  - `any` - Like the `trust` method, but the user name given is ignored. Requires that all databases are configured to log in as a specific user. Additionally, the console database allows any user to log in as admin.
  - `trust` - No authentication is done. The user name must still exist in auth_file.

schnie:
@andriisoldatenko did you find any more information about the `users.txt` file? Seems redundant to have both right?

andriisoldatenko:
@schnie it doesn't work without this file :(
```
PgBouncer needs its own user database. The users are loaded from a text file in the following format:
```
it works with file and without `user= passsword=`

schnie:
@ianstanton @andriisoldatenko so what is the best way forward that works in all situations, typical, and for azure?

andriisoldatenko:
@schnie what do you mean? this PR will work with all databases, i've tested with local postgresql and with Azure instance.

ianstanton:
Closing until testing is complete

ianstanton:
Closing until we do further performance testing with Azure DB PostgreSQL

PR #36, opened 2020-02-24
Decrease PGbouncer logging verbosity

sjmiller609:
A customer reported using these configurations to avoid verbose logging from pgbouncer, which was causing issues for them. I think that connect / disconnect logging is not that helpful in this scenario where there are many components that are constantly connecting and disconnecting, by design. The 'verbose' configuration of 0 is the default for pgbouncer, and is the least verbose.

**Checklist**


schnie:
Looks good, but should we undo the version bump?

PR #39, opened 2020-02-25
Fix 404 for flower sub-path

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/835


andriisoldatenko:
image

schnie:
@andriisoldatenko why did we have to split into two `Ingress` objects?

schnie:
@andriisoldatenko why did we have to split into two `Ingress` objects?

andriisoldatenko:
tl;dr it conflicts with airflow one.

there are some issues:
- https://github.com/kubernetes/ingress-nginx/issues/229
- https://github.com/mher/flower/issues/414

templates/ingress.yaml
```diff
@@ -53,17 +49,67 @@ spec:
...
+    - secretName: ~
+            {{- end }}
+      hosts:
+        - {{- include "deployments_subdomain" . | indent 1 }}
+        - {{- include "flower_subdomain" . | indent 1 }}
+  rules:
+    - host: {{- include "deployments_subdomain" . | indent 1 }}
+      http:
+        paths:
+          - path: /
```
andriisoldatenko: it must be `path: /`

templates/ingress.yaml
```diff
@@ -53,17 +49,67 @@ spec:
...
+{{- end }}
+{{- if and .Values.ingress.enabled (eq .Values.executor "CeleryExecutor") }}
+---
+kind: Ingress
+apiVersion: extensions/v1beta1
+metadata:
+  name: {{ .Release.Name }}-flower-ingress
+  labels:
+    tier: airflow
+    component: airflow-ingress
```
schnie: @andriisoldatenko maybe `flower-ingress` here?

schnie: Also, if we change this to `flower-ingress` do we need do do anything on our Grafana or Prometheus stack? I feel like we may have some stuff that looks for this label, but not sure.

andriisoldatenko: @schnie 
quick look in `github.com/astronomer/astronomer`
```
$ grep --exclude-dir={.git,.idea} -rnw . -e 'airflow-ingress'
```
return nothing.

andriisoldatenko: fixed component name.

andriisoldatenko:
image
image
fixed few bugs at 480d8e2

vishwas-astro:
@schnie can you review and merge please?

PR #40, opened 2020-02-25
Configurable nginx ingress annotations

dimberman:


dimberman:
@schnie I'm thinking that as a stop-gap while we wait for the houston changes, we can simply make all the annotations the defaul value in the values.yaml. Users will have an example of what those annotations can look like, and can easily override it with either an empty string or with their own values.

dimberman:
@schnie @sjmiller609 I've tested locally and ingress renders correctly

values.yaml
```diff
@@ -25,6 +25,26 @@ ingress:
   # Name of tls secret to use on ingress
   tlsSecretName: ~
 
+  # Annotations always injected when configuring Ingress ojbect
+  annotations: |-
```
ashb: What happens if this is not intended correctly (4 spaces, but instead say 2) - does that break the yaml?

If so making these lists, and using `range` in the template is probably safer.

ashb: `Using |-` will strip leading indents anyway, so what happens if this is just set to:

```
    annotations: "kubernetes.io/ingress.class: x"
```

say?

values.yaml
```diff
@@ -25,6 +25,26 @@ ingress:
...
+      kubernetes.io/ingress.class: {{ .Values.ingress.class }}
```
ashb: Given we have an explicit field for this, we should probably leave this one in the template, not in the space for "user" annotations.

dimberman: If we leave it in the template then the user is required to have a value for it. Are there cases where a user would not need this value?

schnie: Yea, the `values.yaml` file does not have templating support since its the input to the actual templating. We may just need to make the corresponding Houston PR.

dimberman: @schnie I was able to do the templating via values.yaml.

```kind: Ingress
apiVersion: extensions/v1beta1
metadata:
  name: release-name-airflow-ingress
  labels:
    tier: airflow
    component: airflow-ingress
    release: release-name
    chart: "airflow-0.12.0-alpha.2"
    heritage: Tiller
  annotations:
    kubernetes.io/ingress.class: nginx
    kubernetes.io/tls-acme: "false"
    nginx.ingress.kubernetes.io/custom-http-errors: "403,404,502,503"
    nginx.ingress.kubernetes.io/auth-url: https://houston./v1/authorization
    nginx.ingress.kubernetes.io/auth-signin: https://app./login
    nginx.ingress.kubernetes.io/auth-response-headers: authorization, username, email
    nginx.ingress.kubernetes.io/proxy-cookie-path: / /release-name/
    nginx.ingress.kubernetes.io/configuration-snippet: |
      if ($host = 'release-name-airflow.' ) {
        rewrite ^ https://deployments.%!s(<nil>)/release-name/airflow permanent;
      }
      if ($host = 'release-name-flower.' ) {
        rewrite ^ https://deployments.%!s(<nil>)/release-name/flower permanent;
      }```

This was generated when enabled=true

Comment (dimberman) [reply values.yaml]: I just had to use the tpl function https://github.com/helm/helm/issues/2133#issuecomment-589695803

Comment (dimberman) [reply values.yaml]: If anything I think these example values offer pretty good documentation.

Comment (schnie) [reply values.yaml]: That output looks a little off. For example the `auth-url` is missing the `baseDomain` part and there are some `<nil>`s in the snippets. Is it because that value isn't set or because of a templating issue?

Comment (dimberman) [reply values.yaml]: @schnie I'm pretty certain it's just because the value isn't set. I'm just using the 'values.yaml' in the repo and not adding anything.

Comment (dimberman) [reply values.yaml]:
```
kind: Ingress
apiVersion: extensions/v1beta1
metadata:
  name: release-name-airflow-ingress
  labels:
    tier: airflow
    component: airflow-ingress
    release: release-name
    chart: "airflow-0.12.0-alpha.2"
    heritage: Tiller
  annotations:
    kubernetes.io/ingress.class: nginx
    kubernetes.io/tls-acme: "false"
    nginx.ingress.kubernetes.io/custom-http-errors: "403,404,502,503"
    nginx.ingress.kubernetes.io/auth-url: https://houston.foobar.com/v1/authorization
    nginx.ingress.kubernetes.io/auth-signin: https://app.foobar.com/login
    nginx.ingress.kubernetes.io/auth-response-headers: authorization, username, email
    nginx.ingress.kubernetes.io/proxy-cookie-path: / /release-name/
    nginx.ingress.kubernetes.io/configuration-snippet: |
      if ($host = 'release-name-airflow.foobar.com' ) {
        rewrite ^ https://deployments.foobar.com/release-name/airflow permanent;
      }
      if ($host = 'release-name-flower.foobar.com' ) {
        rewrite ^ https://deployments.foobar.com/release-name/flower permanent;
      }
spec:
```

This is what it looks like when I added a basedomain

File: values.yaml (hunk 1 above)
Comment (dimberman) [reply values.yaml]: @ashb I'm still not quite certain what the issue is here?

File: values.yaml (hunk 2 above)
Comment (dimberman) [reply values.yaml]: ping @schnie is this concern resolved?

File: values.yaml (hunk 1 above)
Comment (ashb) [reply values.yaml]: By making this a string it's harder to get right -- the indentation matters here more than just for YAML here -- it has to be correct in the end result, once the template is rendered, doesn't it?)

If instead the structure was:

```yaml
  annotations:
       kubernetes.io/ingress.class: "{{ .Values.ingress.class }}"
```

i.e. a dict,  then it's "harder" to get wrong -- if this file is valid YAML the end result when the annotation template is rendered is much more likely to be valid YAML.

dimberman: I just don't think that's flexible enough... we have cases of conditional values, sub-values etc. and I'm not sure we could handle all of those in a dict. Though I guess we could offer both string AND dict? @schnie what do you think?

schnie:
I think I'd prefer the object method instead of string. I think that's how most major other charts do it and it'd be consistent with `.Values.labels`.

dimberman:
@ashb @schnie Is this good to merge?

templates/ingress.yaml
```diff
@@ -16,19 +17,14 @@ metadata:
 {{ toYaml . | indent 4 }}
 {{- end }}
   annotations:
-    kubernetes.io/ingress.class: {{ .Values.ingress.class }}
```
ashb: I Still think we should keep this specific annotation directly here -- we've got a separate value for it, and this change makes that value otherwise unused.

templates/ingress.yaml
```diff
@@ -73,9 +69,6 @@ metadata:
...
-    kubernetes.io/tls-acme: {{ eq .Values.ingress.acme true | quote }}
-    nginx.ingress.kubernetes.io/custom-http-errors: "403,404,502,503"
   {{- if .Values.ingress.auth.enabled }}
     nginx.ingress.kubernetes.io/auth-url: https://houston.{{ .Values.ingress.baseDomain }}/v1/authorization
```
ashb: Shouldn't we remove these like we did for the previous block on L22

templates/ingress.yaml
```diff
@@ -16,19 +17,14 @@ metadata:
 {{ toYaml . | indent 4 }}
 {{- end }}
   annotations:
-    kubernetes.io/ingress.class: {{ .Values.ingress.class }}
-    kubernetes.io/tls-acme: {{ eq .Values.ingress.acme true | quote }}
```
ashb: Likewise here -- removing this annotation makes the `Values.ingress.acme` unused by default.

values.yaml
```diff
@@ -25,6 +25,12 @@ ingress:
   # Name of tls secret to use on ingress
   tlsSecretName: ~
 
+  # Annotations always injected when configuring Ingress ojbect
+  annotations: {}
+
+  # These Annotations will only be injected into the Ingress object if ingress.auth.enabled=true
+  authAnnotations: {}
```
ashb: And follow up question: if this is empty here and controlled on the API/commander side, why do we need a separate value at all?

templates/ingress.yaml
dimberman: @schnie WDYT

values.yaml
dimberman: @ashb we need a separate value because there is an if statement in the chart and if statements don't transfer well to dictionaries.

ashb: Are both of these dictionaries generated from our nodejs app though?

schnie: Yea any `.Value`s that are directly consumed by these annotations can probably be removed. All of this logic can be extracted and moved to the `ingress` section of the config generation here: https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L66-L79.

templates/ingress.yaml
schnie: We may want to move this logic to Houston and remove this special config. Same as my earlier comment below.

dimberman:
@ashb I'm with @schnie here. By giving users (and by the same logic, us) full control of what the annotations section looks like we offer more flexibility. I don't think that a hybrid "some annotations are auto-filled and some are user decided" is ideal

schnie:
We control which versions of the airflow chart are available at the platform level, so we'd release it here, then bump it there whenever we feel ready.

I think we need to follow this up with the Houston PR to compensate for what we're loosing over here. This is becoming super important to our stability and ability to run upgrades. We are fighting an issue right now upgrading cloud due to a similar mismatch that was dropped along the way.

@dimberman can you work on the Houston side of this so we can see both before moving forward too far?

dimberman:
@schnie can do!

dimberman:
@ashb @schnie I've added the flower settings to houston here https://github.com/astronomer/astronomer/pull/579/files. Worth noting that it seems I'm not authorized to set up CI/CD for that project so I think we need to add me somewhere.

values.yaml
```diff
@@ -25,12 +25,17 @@ ingress:
   # Name of tls secret to use on ingress
   tlsSecretName: ~
 
+  # Annotations always injected when configuring webserver Ingress ojbect
+  webserverAnnotations: {}
+
```
schnie:
```suggestion
```

values.yaml
```diff
@@ -25,12 +25,16 @@ ingress:
...
+  # Annotations always injected when configuring Flower Ingress object
+  flowerAnnotations: {}
+
+
```
schnie:
```suggestion
```

andriisoldatenko:
@dimberman  pls resolve conflict

dimberman:
@andriisoldatenko @schnie conflicts resolved

templates/ingress.yaml
```diff
@@ -16,28 +17,8 @@ metadata:
...
-          return 404;
-        }
-        proxy_pass {{ printf "https://%s/%s/airflow/$request_uri" ( include "deployments_subdomain" .) .Release.Name }};
-      }
-    nginx.ingress.kubernetes.io/configuration-snippet: |
-      if ($host = '{{- include "airflow_subdomain" . }}' ) {
-        return 308 {{ printf "https://%s/%s/airflow/$request_uri" ( include "deployments_subdomain" .) .Release.Name }};
-      }
+  {{- range $k, $v := .Values.ingress.webserverAnnotations }}
+    {{ $k | quote }}: {{ tpl $v $ctx | toYaml }}
```
sjmiller609: Please rework this logic to be simpler like just 

{{- .Values.ingress.webserverAnnotations }}

with maybe a wrapping conditional

sjmiller609: it is because some of the string include templating?
```
nginx.ingress.kubernetes.io/auth-signin: https://app.{{ .Values.ingress.baseDomain }}/login
```

sjmiller609: ok I changed my mind.

templates/ingress.yaml
```diff
@@ -16,28 +17,8 @@ metadata:
...
-        if ($host = '{{- include "deployments_subdomain" . }}' ) {
-          return 404;
-        }
-        proxy_pass {{ printf "https://%s/%s/airflow/$request_uri" ( include "deployments_subdomain" .) .Release.Name }};
-      }
-    nginx.ingress.kubernetes.io/configuration-snippet: |
-      if ($host = '{{- include "airflow_subdomain" . }}' ) {
-        return 308 {{ printf "https://%s/%s/airflow/$request_uri" ( include "deployments_subdomain" .) .Release.Name }};
-      }
+  {{- range $k, $v := .Values.ingress.webserverAnnotations }}
```
sjmiller609: how does this template out when webserverAnnotations is empty dict {} or null ~ ?

dimberman: @sjmiller609 thank you for catching that. Go templates FTL :(

sjmiller609: np, we really need something for unit testing helm charts for this exact kind of thing. https://github.com/astronomer/issues/issues/909

dimberman:
@sjmiller609 wouldn't this PR handle that? https://github.com/astronomer/houston-api/pull/243

sjmiller609:
> @sjmiller609 wouldn't this PR handle that? [astronomer/houston-api#243](https://github.com/astronomer/houston-api/pull/243)

Perfect! Thank you.

PR #41, opened 2020-02-25
bump 0.12.0-alpha.3

dimberman:


schnie:
@dimberman I think we'll want to make this to `release-0.12` branch. @sjmiller609 do you agree?

dimberman:
@schnie https://github.com/astronomer/airflow-chart/pull/41

PR #44, opened 2020-02-27
allow keda-operator to access pg_bouncer

dimberman:


templates/pgbouncer/pgbouncer-networkpolicy.yaml
```diff
@@ -36,6 +36,12 @@ spec:
           tier: monitoring
           component: prometheus
           release: {{ .Values.platform.release }}
+    - namespaceSelector:
+        matchLabels:
+          team: operations
+      podSelector:
+        matchLabels:
+          app: keda-operator
```
sjmiller609: I agree with your assessment in slack that this is not secure because there can be other namespaces with this label. I would like to see the following outcome from this PR:

- by default, the minimum network policy is applied to pgbouncer in airflow-chart installation, only allowing access from the workload(s) in this chart that require access.
- There is an optional configuration in values.yaml of type 'list' to add additional ingress rules to the network policy: pgbouncer.networkpolicy.ingress.extraRules

This includes removing the rule to enable prometheus scrape configuration from this chart.

templates/pgbouncer/pgbouncer-networkpolicy.yaml
```diff
@@ -36,6 +36,10 @@ spec:
           tier: monitoring
           component: prometheus
           release: {{ .Values.platform.release }}
+    - namespaceSelector: {}
+      podSelector:
+        matchLabels:
+          app: keda-operator
```
schnie: @dimberman can we use a similar selector scheme as our other stuff?

```
tier: keda
component: operator
release: {{ .Values.platform.release }}
```

dimberman: @schnie I'm not seeing anything in the KEDA helm chart that allows for custom labels :(. I'm gonna try to PR that as it seems like an oversight.

dimberman: https://github.com/kedacore/charts/pull/24/files I've added a PR to the KEDA charts.

dimberman:
@sjmiller609 can we get this merged?

values.yaml
```diff
@@ -139,6 +139,8 @@ workers:
   # Persistence.enabled must be set to false to use KEDA.
   keda:
     enabled: false
+    namespaceLabels: |-
+      name: astronomer
```
schnie: @dimberman this namespace is not guaranteed each time, only when using our terraform.

Also, now that this might be used more generally by the community, it feels like we shouldn't have this pre-populated. Could we pull the default value out here and pass from Houston?

values.yaml
```diff
@@ -272,7 +274,8 @@ statsd:
 pgbouncer:
   # Enable pgbouncer
   enabled: false
-
+  # additional network policies as needed
+  extraNetworkPolicies: ""
```
schnie:
```suggestion

  # Additional network policies as needed
  extraNetworkPolicies: ""

```

schnie: Just some ocd touchups :grimacing:

values.yaml
sjmiller609: Also if you’re going to change this, I prefer dict vs block string and toYaml vs tpl

sjmiller609: Make sure to test empty dict

dimberman: sure

PR #58, opened 2020-03-16
add pathBased enable flag

andriisoldatenko:


andriisoldatenko:
close new fix -> https://github.com/astronomer/airflow-chart/pull/59

PR #60, opened 2020-03-19
Test Local, Kubernetes, and Celery executors in CI

sjmiller609:
Closes: https://github.com/astronomer/issues/issues/978

If CI passes, that indicates these three executors work on Kubernetes 1.14-1.17

### DIFF
--- a/templates/redis/redis-statefulset.yaml
+++ b/templates/redis/redis-statefulset.yaml
@@ -3,7 +3,7 @@
 #################################
 {{- if eq .Values.executor "CeleryExecutor" }}
 kind: StatefulSet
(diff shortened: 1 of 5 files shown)


templates/redis/redis-statefulset.yaml
```diff
@@ -3,7 +3,7 @@
 #################################
 {{- if eq .Values.executor "CeleryExecutor" }}
 kind: StatefulSet
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
```
sjmiller609: this is the only change in the airflow chart for this PR

PR #63, opened 2020-03-31
Do not add user-provided env vars to cleanup job

sjmiller609:


templates/cleanup/cleanup-cronjob.yaml
```diff
@@ -46,6 +46,5 @@ spec:
               # Don't use entry point here, we don't need to wait on pg-bouncer etc being available.
               command: ["airflow-cleanup-pods", "--namespace={{ .Release.Namespace }}"]
               env:
-              {{- include "custom_airflow_environment" . | indent 12 }}
```
andriisoldatenko: nice catch!

PR #64, opened 2020-04-03
Add an initContainer to chown worker logs directory

schnie:
This PR adds an `initContainer` to the workers to prevent a `Permission Denied` issue on boot when running in KinD using the `local-path` provisioner.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -161,6 +161,10 @@ workers:
     size: 100Gi
     # If using a custom storageClass, pass name ref to all statefulSets here
     storageClassName:
+    # Execute init container to chown log directory
+    # This is currently only needed in KinD, due to usage
+    # of local-path provisioner.
+    fixPermissions: false
 
   resources: {}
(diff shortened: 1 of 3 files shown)


schnie:
Still not sure why this doesn't happen in Circle though. Any ideas?

andriisoldatenko:
@schnie rerun fixed the flaky problem!
LGTM

values.yaml
```diff
@@ -161,6 +161,10 @@ workers:
     size: 100Gi
     # If using a custom storageClass, pass name ref to all statefulSets here
     storageClassName:
+    # Execute init container to chown log directory
```
schnie, line 164:
```suggestion
    # Execute init container to chown log directory.
```

schnie:
> @schnie rerun fixed the flaky problem!
> LGTM

I was wondering why we don't experience this problem in Circle though. I had to add this to get it to actually boot up in kind.

PR #66, opened 2020-04-06
Clean up platform references

schnie:
This PR removes the last remaining references to anything that deals with "the platform", "astronomer", "astro", or our surrounding infrastructure in general.


schnie:
Here is the `values.yaml` I was using to recreate the same output as I pulled each part out. This represents all (some already exist but required to test in isolation) the values that will either need to be provided statically to every deployment via Houston's `ConfigMap`, or be generated dynamically by Houston. Those PRs will be opened next.

```
workers:
  persistence:
    fixPermissions: true

executor: CeleryExecutor

networkPolicies:
  enabled: true

labels:
  platform: "astronomer"
  workspace: "123456789"

webserver:
  extraNetworkPolicies:
  - namespaceSelector: {}
    podSelector:
      matchLabels:
        tier: nginx
        component: ingress-controller
        release: release-name
  webserverConfig: |
    import os
    from airflow import configuration as conf
    from flask_appbuilder.security.manager import AUTH_REMOTE_USER
    basedir = os.path.abspath(os.path.dirname(__file__))

    # The SQLAlchemy connection string.
    SQLALCHEMY_DATABASE_URI = conf.get('core', 'SQL_ALCHEMY_CONN')

    # Flask-WTF flag for CSRF
    CSRF_ENABLED = True

    # ----------------------------------------------------
    # AUTHENTICATION CONFIG
    # ----------------------------------------------------
    AUTH_TYPE = AUTH_REMOTE_USER

    from astronomer.flask_appbuilder.security import AirflowAstroSecurityManager
    SECURITY_MANAGER_CLASS = AirflowAstroSecurityManager

  extraVolumes:
  - name: signing-certificate
    secret:
      secretName: "jwtSigningCertificateSecretName"

  extraVolumeMounts:
  - name: signing-certificate
    mountPath: "/etc/airflow/tls"
    readOnly: true

flower:
  extraNetworkPolicies:
  - namespaceSelector: {}
    podSelector:
      matchLabels:
        tier: nginx
        component: ingress-controller
        release: release-name

statsd:
  service:
    extraAnnotations:
      astronomer.io/platform-release: release-name

  extraNetworkPolicies:
  - namespaceSelector: {}
    podSelector:
      matchLabels:
        tier: monitoring
        component: prometheus
        release: release-name

pgbouncer:
  enabled: true

  service:
    extraAnnotations:
      astronomer.io/platform-release: release-name

  extraNetworkPolicies:
  - namespaceSelector: {}
    podSelector:
      matchLabels:
        tier: monitoring
        component: prometheus
        release: release-name

env:
  - name: ASTRONOMER__AIRFLOW__WORKER_LOG_RETENTION
    value: "3"
  - name: ASTRONOMER__JWT_SIGNING_CERT
    value: "/etc/airflow/tls/tls.crt"
  - name: ASTRONOMER__JWT_AUDIENCE
    value: "deployments.baseDomain"
```

templates/flower/flower-networkpolicy.yaml
```diff
@@ -23,17 +23,12 @@ spec:
...
-{{- if .Values.platform.release }}
   ingress:
   - from:
-    - namespaceSelector: {}
-      podSelector:
-        matchLabels:
-          tier: nginx
-          component: ingress-controller
-          release: {{ .Values.platform.release }}
+{{- if .Values.flower.extraNetworkPolicies }}
```
ashb, line 28: Does this render right (and work) when extraNetworkPolicies is empty?

templates/pgbouncer/pgbouncer-networkpolicy.yaml
```diff
@@ -44,19 +44,9 @@ spec:
 {{- if .Values.pgbouncer.extraNetworkPolicies}}
 {{ toYaml .Values.pgbouncer.extraNetworkPolicies | indent 4}}
 {{- end}}
-    {{- if .Values.platform.release }}
-    - namespaceSelector: {}
```
ashb, line 48: How are we applying this one still? I don't see any way to add this back in?

schnie:
We still point to all our images for the components. Debating what to do there. Any ideas?

templates/webserver/webserver-deployment.yaml
```diff
@@ -83,10 +83,8 @@ spec:
               mountPath: {{ template "airflow_local_setting_path" . }}
               subPath: airflow_local_settings.py
               readOnly: true
-{{- if .Values.webserver.jwtSigningCertificateSecretName }}
-            - name: signing-certificate
-              mountPath: {{ template "airflow_webserver_jwt_cert_dir" . }}
-              readOnly: true
+{{- if .Values.webserver.extraVolumeMounts }}
```
ashb, line 86: We're still mounting webserver_config.py on line 80, should we be?

Equally for airflow_local_settings.py - that's probably a platform only-feature.

values.yaml
```diff
@@ -394,29 +416,22 @@ podMutation:
...
-#                  - "true"
+  # affinity: {}
+  #   nodeAffinity:
+  #     requiredDuringSchedulingIgnoredDuringExecution:
+  #       nodeSelectorTerms:
+  #       - matchExpressions:
+  #         - key: "company.io/dynamic-pods"
+  #           operator: In
+  #           values:
+  #           - "true"
```
sjmiller609, line 427: Is the top line supposed to be commented out?

values.yaml
```diff
@@ -255,8 +251,26 @@ webserver:
     lastName: user
     password: admin
 
+  # Mount additional volumes into webserver.
+  extraVolumes: []
+  extraVolumeMounts: []
+
+  # This will be mounted into the Airflow Webserver as
+  # a custom webserver_config.py
+  webserverConfig: ~
```
ashb, line 260: Hmmm, I wonder if it is possible to get the feature we want without having this as a field in the helm chart -- it's a bit odd to have a "large" block like this, and most people would be able to bake the webserver_config.py they need in to their images.

templates/flower/flower-networkpolicy.yaml
schnie: I think so. I figured we could leave the ports configured regardless and optionally let users allow custom pods to communicate. So the default is that there are no sources of traffic allowed.

```# Source: airflow/templates/flower/flower-networkpolicy.yaml
## Airflow Flower NetworkPolicy
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: kindly-labradoodle-flower-policy
  labels:
    tier: airflow
    component: airflow-flower-policy
    release: kindly-labradoodle
    chart: "airflow-0.12.0-alpha.5"
    heritage: Tiller
    platform: astronomer
    workspace: "123456789"
    
spec:
  podSelector:
    matchLabels:
      tier: airflow
      component: flower
      release: kindly-labradoodle
  policyTypes:
  - Ingress
  ingress:
  - from:
    ports:
    - protocol: TCP
      port: 5555
```

templates/configmap.yaml
```diff
@@ -92,36 +92,14 @@ data:
...
-    # ----------------------------------------------------
-{{- if .Values.webserver.jwtSigningCertificateSecretName }}
-    AUTH_TYPE = AUTH_REMOTE_USER
-
-    from astronomer.flask_appbuilder.security import AirflowAstroSecurityManager
-    SECURITY_MANAGER_CLASS = AirflowAstroSecurityManager
+    {{- .Values.webserver.webserverConfig | nindent 4 }}
 {{- end }}
 
 {{- if .Values.scheduler.airflowLocalSettings }}
```
kaxil, line 105: We may want to think about removing our `pod_mutation_hook` settings, we can pass `airflow_local_settings` from Astro chart probably ?

kaxil: https://github.com/astronomer/airflow-chart/pull/66/files#diff-1d2f4c201ce0c92586e4e3c5f9a6e03eR109-R130

templates/flower/flower-networkpolicy.yaml
schnie: Well, good point. It renders ok, but it's actually allow traffic from anyone, so I'll have to fix that.

```
$ kc describe networkpolicy intentional-lizard-flower-policy 
Name:         intentional-lizard-flower-policy
Namespace:    default
Created on:   2020-04-08 13:13:29 -0400 EDT
Labels:       chart=airflow-0.12.0-alpha.5
              component=airflow-flower-policy
              heritage=Tiller
              platform=astronomer
              release=intentional-lizard
              tier=airflow
              workspace=123456789
Annotations:  <none>
Spec:
  PodSelector:     component=flower,release=intentional-lizard,tier=airflow
  Allowing ingress traffic:
    To Port: 5555/TCP
    From: <any> (traffic not restricted by source)
  Not affecting egress traffic
  Policy Types: Ingress
```

templates/pgbouncer/pgbouncer-networkpolicy.yaml
schnie: Good eye, but right above this @dimberman already added the `extraNetworkPolicies` chunk for KEKA support. I essentially copied this chunk to the other `NetworkPolicies` we had to tweak. So this change is just removing the part that was hard-coded to `platform.release`. We'll just have to move this up to Houston somehwere, so made a note of it in my `values.yaml` I referenced in a comment here.

templates/flower/flower-networkpolicy.yaml
schnie: Well, now that I think about this again. What would we want open source default behavior to be? Should flower and the webserver allow traffic from anyone by default? Maybe that's preferable as default. :thinking:

templates/webserver/webserver-deployment.yaml
schnie: We actually have a default `airflow_local_settings.py` baked in that adds `kubernetes-executor` / `kubernetes-pod-operator` labels to dynamic pods, which is kind of useful, and kind of serves as a nice little demo of the feature.

I could see us just removing it and treating it just like we're treating the `webserver_config.py` though. 

Yes I think you are right about the `webserver_config.py`. That should be conditionally added, if one is defined. I'll fix that.

values.yaml
schnie: Yea I agree its a bit awkward. I guess it could technically be supported using the `volume` mount options right above it, assuming they create the file in their own `ConfigMap` or `Secret`.

templates/flower/flower-networkpolicy.yaml
andriisoldatenko: >Should flower and the webserver allow traffic from anyone by default? 
+1 for OSS chart.

values.yaml
schnie: Ah, good catch. Fixed.

templates/configmap.yaml
schnie: I'm fine with removing it, maybe leave something in there commented out as an example? I do feel like this specific feature is being used in the wild a lot, so it might be nice to have a little example in there.

kaxil: Yeah I agree., something like:

```yaml
# 1) A "policy" function that has the ability to mutate task attributes based on other task or DAG attributes.
# 2) "pod_mutation_hook". This could be used, for instance, to add sidecar or init containers
#     to every worker pod or just add labels to all pods launched by KubernetesExecutor or KubernetesPodOperator.

#     def pod_mutation_hook(pod: Pod):
#         extra_labels = {
#             "kubernetes-executor": "False",
#             "kubernetes-pod-operator": "False"
#         }
#         if 'airflow-worker' in pod.labels.keys() or \
#                 conf.get('core', 'EXECUTOR') == "KubernetesExecutor":
#             extra_labels["kubernetes-executor"] = "True"
#         else:
#             extra_labels["kubernetes-pod-operator"] = "True"
#         pod.labels.update(extra_labels)
```

templates/flower/flower-networkpolicy.yaml
schnie: I think we can leave it as is for now then. By default it'll be open, and as we apply policies, it'll get locked down.

templates/configmap.yaml
kaxil: What do you think about the above command

templates/webserver/webserver-deployment.yaml
schnie: fixed these.

templates/configmap.yaml
```diff
@@ -92,61 +92,17 @@ data:
...
     release = {{ .Release.Name }}
-    platform = {{ .Values.platform.release }}
-    workspace = {{ .Values.platform.workspace }}
-{{- end }}
-
-    [astronomer]
-{{- if .Values.webserver.jwtSigningCertificateSecretName }}
-    jwt_signing_cert = {{ template "airflow_webserver_jwt_cert_path" . }}
-    jwt_audience = {{ printf "%s/%s" ( include "deployments_subdomain" . ) .Release.Name }}
+    {{- range $key, $value := .Values.labels }}
```
andriisoldatenko, line 95: did you have a chance to test it with in real platform/cluster? does it practically work?

schnie:
Astronomer PR opened here: https://github.com/astronomer/astronomer/pull/666

ashb:
@schnie Do we care about this PR anymore?

schnie:
Probably not. I think we've effectively decided to support the astro version of the chart for the rest of the platforms lifetime. For the new stuff, were moving onto the operator.

PR #68, opened 2020-04-10
Wait for DB migration with the right image

sjmiller609:
issue when 'astro deploy' new airflow version: webserver does not start

### DIFF
--- a/templates/webserver/webserver-deployment.yaml
+++ b/templates/webserver/webserver-deployment.yaml
@@ -57,7 +57,7 @@ spec:
       {{- end }}
       initContainers:
         - name: wait-for-airflow-migrations
-          image: {{ template "default_airflow_image" . }}
+          image: {{ template "airflow_image" . }}
           imagePullPolicy: {{ .Values.images.airflow.pullPolicy }}
           args: ["airflow-migration-spinner", "--timeout=60"]
           env:


schnie:
What causes this to not work? It uses that default image to avoid the requirement of the user having that script baked into their iamge. In case someone hasn't upgraded the image in a bit. Might not be a problem anymore though.

sjmiller609:
in practice on 1.10.10, stuck in init on webserver

schnie:
All our modern builds have this script baked in, but some deployments might not be running the latest image and a code push could potentially cause it to reboot and error out on the `initContainer`.

Not sure of a good way to determine the impact of this.

sjmiller609:
Example:

Customer wants to upgrade prod and not update their airflow version https://astronomerteam.slack.com/archives/CLH8HNDCZ/p1586894902073000?thread_ts=1586875655.070100&cid=CLH8HNDCZ

In a replicated environment, we have found webserver stuck init 0:1

```
cometary-cosmic-5156-webserver-955497f45-wmd6l    0/1     Init:0/1                0  
```

sjmiller609:
In the above case, my suggestion will not work because airflow-migration-spinner does not exist. It remains for me to figure out why it's stuck, let me check

sjmiller609:
yeah I think it's an unrelated reason (fixed by this PR) https://github.com/astronomer/astronomer/pull/645/files

sjmiller609:
On deeper inspection, I believe that the script anticipates the version of airflow, which makes using the default image incompatible with any other airflow version. https://github.com/astronomer/astronomer-airflow-scripts/blob/d82dc55647c35c1d163d043b80b3dddce968455a/astronomer/migration_spinner/command_line.py#L15-L38

we need to support multiple airflow versions because https://github.com/astronomer/issues/issues/1132

schnie:
Ok that makes sense, but if someone pushes an older image up this will fail and the only option for them will be to get onto a supported release.

sjmiller609:
> Ok that makes sense, but if someone pushes an older image up this will fail and the only option for them will be to get onto a supported release.

yeah I think so. but I would be curious if any version other than AC 1.10.7 works on airflow-chart 0.12.2, because I have not found one.

schnie:
The latest versions of all `ap-airflow` images should have the script baked in - https://github.com/astronomer/ap-airflow/blob/master/1.10.5/buster/Dockerfile#L119.

Since some of those tags are moving/non-stable we can't currently guarantee the presence of the script. If we do hit this with a customer, and their `Dockerfile` tag looks normal, then maybe a fresh build without caching will get it through.

As mentioned in dm, if we knew the version of Airflow that deployment was running, Houston could dynamically pass in an appropriate image tag that we had complete control over.

@vishwas-astro this is another thing that we could control much better if we had airflow-version-awareness baked into the platform. This gets more important by the day it seems.

PR #69, opened 2020-04-10
Fixes bug where imagePullPolicy is not universally acknowledged

dimberman:


dimberman:
This bug prevents us from using the helm chart for CI/CD

PR #73, opened 2020-04-15
user-provided image for DB migration wait

sjmiller609:
forgot this part in https://github.com/astronomer/airflow-chart/pull/68


sjmiller609:
ty

PR #74, opened 2020-04-29
Adds TargetQueryValue field to KEDA ScaledObject

dimberman:
Newer versions of KEDA require the TargetQueryValue to be filled.
I've defaulted it to one so that autoscaling will always happen

### DIFF
--- a/templates/workers/worker-kedaautoscaler.yaml
+++ b/templates/workers/worker-kedaautoscaler.yaml
@@ -26,5 +26,6 @@ spec:
     - type: postgresql
       metadata:
         connection: AIRFLOW_CONN_AIRFLOW_DB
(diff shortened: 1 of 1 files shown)


dimberman:
@schnie @sjmiller609 this fix is required for KEDA otherwise we would need to go in and manually edit the scaledObjects. Everything else we were able to handle via values.yaml

schnie:
@dimberman what does `TargetQueryValue` doing under the hood?

Is this something we'd want to change dynamically in our system for any reason?

dimberman:
@schnie I haven never had to set it to anything but 1.

dimberman:
Also if we can hotfix this int 0.13 that would be ideal (but 100% this needs to be in 0.14)

PR #75, opened 2020-04-30
Add networkPolicy for keda-apiserver

dimberman:
In new versions of KEDA, the apiserver is a seperate pod
with a different label. This PR adds a networkPolicy that
allows the KEDA apiserver to communicate with pgbouncer

### DIFF
--- a/templates/pgbouncer/pgbouncer-networkpolicy.yaml
+++ b/templates/pgbouncer/pgbouncer-networkpolicy.yaml
@@ -40,6 +40,16 @@ spec:
 {{- end }}
         matchLabels:
           app: keda-operator
+{{- if .Values.workers.keda.namespaceLabels }}
+    - namespaceSelector:
+       matchLabels:
+{{ toYaml .Values.workers.keda.namespaceLabels | indent 10}}
+      podSelector:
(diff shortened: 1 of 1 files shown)


templates/pgbouncer/pgbouncer-networkpolicy.yaml
```diff
@@ -40,6 +40,16 @@ spec:
 {{- end }}
         matchLabels:
           app: keda-operator
+{{- if .Values.workers.keda.namespaceLabels }}
```
ashb, line 43: Line 33-42 is almost exactly a duplicate of this. Meaning if `.Values.workers.keda.namespaceLabels` is set we'd end up with them twice in the output. That doesn't seem right.

dimberman: It should be the same because the operator and the apiserver should never live in different namespaces. Up until this version these were actually the same pod, so they're very tightly coupled.

ashb: Ohh, this is two separate selectors. Gotcha.

PR #77, opened 2020-05-05
KEDA maxReplicaCount matches celery worker replicas

dimberman:
We want to bill based on the maximum *POSSIBLE* workers, so if a user
turns on KEDA autoscaling, they should still be limited based on their
billing


dimberman:
@ryw maxReplicaCount was the only relevant one I could find. Could you please double-check and see if there's anything else you'd like me to change?

PR #79, opened 2020-05-07
Add extra labels and annotations to task pods

juniorz:
It is common to the k8s ecosystem having add-ons that are configurable based on pod's annotations/labels (for example, [Istio sidecar injection](https://istio.io/docs/ops/configuration/mesh/injection-concepts/) and [Vault secrets injection](https://www.hashicorp.com/blog/injecting-vault-secrets-into-kubernetes-pods-via-a-sidecar/)).

This PR makes it possible for pods spun up using `KubernetesExecutor` or `KubernetesPodOperator` to better integrate with such add-ons.

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -147,6 +147,8 @@ data:
             extra_labels["kubernetes-pod-operator"] = "True"
 
         pod.labels.update(extra_labels)
+        pod.labels.update({{ toJson .Values.podMutation.labels }})
+        pod.annotations.update({{ toJson .Values.podMutation.annotations }})
         pod.tolerations += {{ toJson .Values.podMutation.tolerations }}
         pod.affinity.update({{ toJson .Values.podMutation.affinity }})
 {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -404,6 +404,12 @@ podMutation:
 #                values:
 #                  - "true"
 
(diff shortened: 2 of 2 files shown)


juniorz:
cc/ @dimberman

juniorz:
@morhook , thank you for reviewing this PR!
Let me know if there's anything missing from my side to get this merged!

schnie:
Hey @juniorz thanks for the PR! I'd be happy to merge this right in, but we have a larger PR in the works that actually removes this file entirely from the default configuration (along with a lot of other things). 

This is in an effort to make this chart more agnostic and generally usable outside of Astronomer. That work is being down over on this PR https://github.com/astronomer/airflow-chart/pull/66. A similar draft PR is open on the apache/airflow repo to add this as the official chart - https://github.com/apache/airflow/pull/8777.

I think we may have discussed this file in particular in the PR, but we decided to make it empty by default and just have an example commented out that folks could copy / paste for an example. I think use-cases of this pod mutation hook are very different from org to org so it made sense to just give full control to users.

Interested to hear what you think about this.

juniorz:
@schnie , thanks for the prompt feedback!

I believe we will just keep defining our own "mutation hook" (via `scheduler.airflowLocalSettings`), and re-evaluate where to go once the new version is released.

PR #80, opened 2020-05-07
Make configurable deletion of pods

morhook:
While developing with my minikube (and KubernetesExecutor) some DAGs, I've realized deleting the pods always makes a little bit hard to guess what happend on that Task/DAG run. 

Adding this as a values.yaml parameter


schnie:
Hey @morhook thanks for the PR!

What do you think about passing this override in via the `env` section? Mostly in an effort to not have to map a 1:1 airflow setting to chart setting. Would that work for you? Env vars should override anything set in this file and allow you to set _any_ airflow setting.

morhook:
Not sure how that will work, but if it makes sense to you maybe you can
explain it better with an alternate PR?

morhook:
After thinking what you said, I have realized you meant a code like this:

```yaml
  env:
    - name: AIRFLOW__KUBERNETES__DELETE_WORKER_PODS
      value: "False"
```

Closing PR and Thanks!

PR #84, opened 2020-05-22
Delete then add update plan for scheduler

sjmiller609:
- delete before replace update strategy for scheduler
- remove pdb: not applicable to workloads with 1 replica


jpweber:
I understand removing the pdb, but I don't understand what this change is for regarding adding the rolling update strategy part. What is the context of this change?

templates/scheduler/scheduler-deployment.yaml
```diff
@@ -32,6 +32,10 @@ spec:
   serviceName: {{ .Release.Name }}-scheduler
 {{- end }}
   replicas: 1
+  strategy:
+    type: RollingUpdate
```
jpweber, line 36: The type here needs to be `Recreate`, not rolling update. 

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#writing-a-deployment-spec

PR #85, opened 2020-05-28
Fix duplicate labels "kubernetes-executor" and "kubernetes-pod-operator"

morhook:
After some testing on my minikube and my company infrastructure, we've
found out that the label "kubernetes-executor" (created by this code) usually appears also with
"kubernetes_executor" (created by airflow code). This produces a breakup on our label processing
in prometheus/metrics.

Also detected same anti-pattern on label "kubernetes-pod-operator" (and
"kubernetes_pod_operator"). 

I think it makes sense for this code to be changed to avoid this
duplication.

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -136,15 +136,15 @@ data:
     def pod_mutation_hook(pod: Pod):
 
         extra_labels = {
-            "kubernetes-executor": "False",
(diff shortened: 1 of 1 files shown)


schnie:
Hey @morhook great catch. We discovered this same issue in our Cloud platform but handled it via our Prometheus configuration but this will remove it at the source. Thanks!

PR #86, opened 2020-06-02
Use new matrix builds Circle CI feature

sjmiller609:
also, add kube 1.18 to tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,9 +64,6 @@ jobs:
       kube_version:
         type: string
         default: "1.18.2"
-    environment:
-      KUBE_VERSION: 1.14.10
-      EXECUTOR: "LocalExecutor"
     steps:
       - attach_workspace:
           at: /tmp/workspace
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -94,113 +54,29 @@ jobs:
...
+    parameters:
+      executor:
+        type: string
+        default: "LocalExecutor"
+      kube_version:
+        type: string
+        default: "1.18.2"
     environment:
       KUBE_VERSION: 1.14.10
       EXECUTOR: "LocalExecutor"
```
kaxil, line 69: We can remove this now I think

sjmiller609: oops, good catch

kaxil: or use parameters here:

```
```suggestion
      KUBE_VERSION: << parameters.kube_version >>
      EXECUTOR: << parameters.executor >>
```

sjmiller609: idk if that would work or not though

PR #95, opened 2020-06-25
Fix typo of resultBackendConnection

vsimon:


vsimon:
I just saw the apache/airflow chart so should everything be merged into there instead and cease here? 

This chart can be installed via a "repo" and that other one seems to install only from source? Until it is publish-ready, what is the recommended route to take?

schnie:
Hey @vsimon I'd probably recommend installing the apache version from source. It is a little bit more friendly to the community needs, as opposed to the Astronomer needs.

In addition to getting the community to converge on a single Kube deployment, this chart was getting pulled in too many directions, so splitting it out should let them evolve quicker.

Ideally we can get the apache version published somewhere soon.

vsimon:
OK, thanks @schnie 

I put this up there too https://github.com/apache/airflow/pull/9537

schnie:
Thanks!

PR #97, opened 2020-07-06
Make Scheduler livenessProbe HA ready.

ashb:
To support running with more than a single scheduler pod we can no
longer rely on `most_recent_job` -- as that would simply select the most
recent row to have beaten, which could be from a different pod.


schnie:
Cool, looks like this should continue to work for existing deployments right?

ashb:
> Cool, looks like this should continue to work for existing deployments right?

Yup.

PR #100, opened 2020-07-14
Add code to be able to add annotations to create-user-job

fjbotto:
Our company is using Istio, this is making the _create-user-job_ not being able to end, and also crashing our CD. In order to fix this we are adding a parameter to be able to add an annotation to that job so we can disable the istio sidecar injection.

Issue reported:

- Istio: https://github.com/istio/istio/issues/11045 and https://github.com/istio/istio/issues/6324
- Kubernetes: https://github.com/kubernetes/kubernetes/issues/65502

More details on the issue:

- https://medium.com/redbox-techblog/handling-istio-sidecars-in-kubernetes-jobs-c392661c4af7

### How to lint the change:

Add "sidecar.istio.io/inject": "False" to createUserJobAnnotations

Execute `helm template .`  and search for ## Airflow Create User Job the new annotation added.

### DIFF
--- a/requirements.lock
+++ b/requirements.lock
@@ -2,5 +2,5 @@ dependencies:
 - name: postgresql
   repository: https://kubernetes-charts.storage.googleapis.com
   version: 6.3.12
-digest: sha256:1750ddcc948f15716d157d6a854b59e37571473b2a3390a3673b224b71a56308
-generated: "2019-10-24T16:03:26.569269284-04:00"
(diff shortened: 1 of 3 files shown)


fjbotto:
Can someone with permissions re-run the failing job (_airflow-test-KubernetesExecutor-1.15.7_)?

fjbotto:
Can someone review it?

PR #101, opened 2020-07-15
add server tls sslmode

andriisoldatenko:
If server requires TLS and rejects plain TCP connection, we need ability to configure it.

### DIFF
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -160,6 +160,7 @@
 [pgbouncer]
 pool_mode = transaction
 listen_port = {{ .Values.ports.pgbouncer }}
+server_tls_sslmode = {{ .Values.pgbouncer.serverTlsSslmode }}
 listen_addr = *
 auth_type = md5
 auth_file = /etc/pgbouncer/users.txt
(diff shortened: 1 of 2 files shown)


andriisoldatenko:
practically works with enabled SSL.

templates/_helpers.yaml
```diff
@@ -160,6 +160,7 @@
 [pgbouncer]
 pool_mode = transaction
 listen_port = {{ .Values.ports.pgbouncer }}
+server_tls_sslmode = {{ .Values.pgbouncer.serverTlsSslmode }}
```
schnie, line 163: @andriisoldatenko do we may need to add a `.Values.pgbouncer.serverTlsSslmode | default "prefer"` here to support a clean upgrade from a version of the chart that does not include this value.

You and @sjmiller609 did some work here so not sure if we still need to do this or not. What do you think?

andriisoldatenko: @schnie so do you think this won't work?
https://github.com/astronomer/airflow-chart/pull/101/files#diff-ba05023d0f1b5d4b64b6cad4cf9a7ecdR313

andriisoldatenko:
@schnie PTAL

PR #104, opened 2020-07-17
Fix HA-ready scheduler probes

sjmiller609:
Closes: https://github.com/astronomer/issues/issues/1560

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,7 @@ jobs:
   airflow-test:
     machine:
       image: ubuntu-1604:201903-01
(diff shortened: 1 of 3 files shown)


sjmiller609:
I think the readiness probe is too slow to pass CI

sjmiller609:
it it passing locally, so I increased the resources allocated in CI and set an initial delay for probes.

sjmiller609:
that did the trick. I think it was just a race condition on account of low resources. the CI instance size is doubled, but instead of taking ~6m, now it takes 4m, so this is a small cost increase.

PR #106, opened 2020-07-28
Don't use custom environment variables in init containers

sjmiller609:
This is follow up from a customer upgrade where custom environmet variables in the Dockerfile of the user-provided container (custom airflow config for airflow secrets backend) caused issues in the database migration (or wait for migration) steps.


sjmiller609:
I'm trying to understand if there is a use case for having custom vars just for the init containers

sjmiller609:
There is an option for me to add new helm variables env_db_migration or something

PR #113, opened 2020-08-14
Fix/1409 secret synchronization

kushalmalani:
Adding secretRef to add secrets to a pod. Fix has multi helper functions and a toggle to make sure we are able to go back to the old way of adding secrets if needed.

Used local kind cluster and helm upgrade commands to verify the toggle functionality


templates/_helpers.yaml
```diff
@@ -53,15 +53,21 @@
...
 {{- define "custom_airflow_environment" }}
-  # Dynamically created environment variables
-  {{- range $i, $config := .Values.env }}
-  - name: {{ $config.name }}
-    value: {{ $config.value | quote }}
-    {{- if eq $.Values.executor "KubernetesExecutor" }}
-  - name: AIRFLOW__KUBERNETES_ENVIRONMENT_VARIABLES__{{ $config.name }}
-    value: {{ $config.value | quote }}
-    {{- end }}
+  {{- if .Values.mountAllFromSecretName }}
```
kushalmalani, line 56: @schnie Is there a way we can test the if condition with the secretName equals {{ .Release.Name }}-env inside the if condition? I tried that, but helm doesnt like it.

schnie: I think this is fine. Houston will take care of passing the right format in. Should be able to trust that. Plus it leaves the chart "generic", so someone else could put whatever they want there if they pick it up for AC purposes.

templates/_helpers.yaml
```diff
@@ -70,12 +76,30 @@
...
+{{- end }}
+
+{{- define "custom_airflow_environment_config" }}
+  # Dynamically created environment variables
+  {{- range $i, $config := .Values.env }}
+  - name: {{ $config.name }}
+    value: {{ $config.value | quote }}
+    {{- if eq $.Values.executor "KubernetesExecutor" }}
+    - name: AIRFLOW__KUBERNETES_ENVIRONMENT_VARIABLES__{{ $config.name }}
+      value: {{ $config.value | quote }}
```
schnie, line 88: @kushalmalani should these be un-indented to line up with the ones right above it?

kushalmalani: oh yeah! Copy pasting in VS Yaml is horrible. Good catch! I will get this in and fix the test cases

kushalmalani:
@schnie Do we need to bump up the Chart version? It seems outdated compared to the tags i see on github

templates/_helpers.yaml
```diff
@@ -70,12 +76,30 @@
...
+  - name: AIRFLOW__KUBERNETES_ENVIRONMENT_VARIABLES__{{ $config.name }}
+    value: {{ $config.value | quote }}
+    {{- end }}
+  {{- end }}
   {{- if eq .Values.executor "KubernetesExecutor" }}
     {{- range $i, $config := .Values.secret }}
-  - name: AIRFLOW__KUBERNETES_SECRETS__{{ $config.envName }}
-    value: {{ printf "%s=%s" $config.secretName $config.secretKey }}
+    - name: AIRFLOW__KUBERNETES_SECRETS__{{ $config.envName }}
+      value: {{ printf "%s=%s" $config.secretName $config.secretKey }}
```
schnie, line 94: @kushalmalani is this indentation right?

kushalmalani: Yeah, i was able to add those env vars to the pods and helm commands didnt throw any errors for that.

templates/_helpers.yaml
```diff
@@ -53,6 +53,32 @@
 
 {{/* User defined Airflow environment variables */}}
 {{- define "custom_airflow_environment" }}
+{{- if .Values.mountAllFromSecretName }}
+envFrom:
+{{- include "custom_airflow_environment_mount_all" . }}
+env:
+{{- include "custom_airflow_environment_config" . }}
+{{- include "standard_airflow_environment" . }}
+{{- else }}
```
schnie, line 62: @kushalmalani does this work right? I thought the last time we talked we determined that one of these - `env` or `envFrom` will clobber the other.

Also, if we are using `mountAllFromSecretName`, we don't want both `custom_airflow_environment_mount_all` and `custom_airflow_environment_config` right? It's the new vs old way of templating the user-supplied env vars in right?

kushalmalani: @schnie It does work right. The reason for separating them was to make sure it works fine and it only clobbers if we are passing the same env vars via envFrom and env, in which case env takes precedence.

The old way of adding custom env vars is defined in "custom_airflow_environment_mount_single". The config part is common, which adds KubernetesExecutor related environment variables.

templates/secrets/custom-airflow-secret.yaml
```diff
@@ -0,0 +1,22 @@
...
+{{- with .Values.labels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+    "helm.sh/hook-weight": "0"
+type: Opaque
+data:
+{{- end }}
```
andriisoldatenko, line 22:
```suggestion
{{- end }}

```

kushalmalani: Will add it in a single commit if there are more changes.

templates/_helpers.yaml
schnie: Oh I see. Should we rename that helper to be kube executor specific to make that clear?

Also, instead of iterating over all the env vars for the kube executor section, I wonder if there is a way to use the `envFrom` syntax. Otherwise we run into similar issues... :thinking:

schnie: @dimberman know if thats possible?

kushalmalani: @schnie That helper also adds env vars from .Values.env for non-kube executor part. I could pull that out into a separate helper, but i felt i was maybe over engineering it.

I wondered if i could use envFrom instead but that requires us to give a secret name for it to pull values from and KubernetesExecutor env vars are created with that range. We will need to create a separate secret file for that.

templates/_helpers.yaml
```diff
@@ -70,12 +76,24 @@
         name: {{ $config.secretName }}
         key: {{ default "value" $config.secretKey }}
   {{- end }}
-  {{- if eq .Values.executor "KubernetesExecutor" }}
-    {{- range $i, $config := .Values.secret }}
-  - name: AIRFLOW__KUBERNETES_SECRETS__{{ $config.envName }}
-    value: {{ printf "%s=%s" $config.secretName $config.secretKey }}
```
schnie, line 76: @kushalmalani we don't need to delete this chunk right? If someone is using the old method with `.Values.secret`, they'd still want those envs to get set into the dynamically launched worker pods.

templates/secrets/custom-airflow-secret.yaml
```diff
@@ -0,0 +1,22 @@
+################################
```
schnie, line 1: @kushalmalani we shouldn't need this since Commander will take care of creating the secret out-of-band from the helm release cycle. Was this for testing purposes?

templates/_helpers.yaml
```diff
@@ -53,15 +53,21 @@
 
 {{/* User defined Airflow environment variables */}}
 {{- define "custom_airflow_environment" }}
```
schnie, line 55: @kushalmalani since this has taken on some new meaning what if we named this helper just `airflow_environment` or something, since it now contains custom envs, as well as the standard envs.

templates/_helpers.yaml
```diff
@@ -53,15 +53,21 @@
...
+{{- include "standard_airflow_environment" . }}
+{{- else }}
+env:
+{{- include "custom_airflow_environment_mount_single" . }}
+{{- include "custom_airflow_environment_config" . }}
+{{- include "standard_airflow_environment" . }}
+{{- end -}}
+{{- end -}}
+
+{{- define "custom_airflow_environment_mount_single" }}
```
schnie, line 70: @kushalmalani what do you think about combining the parts that loop and combine `.Values.secret` and `.Values.env` into a single helper called `custom_airflow_environment_mount_individual`? Slightly different, but maybe a little clearer to the reader?

This would essentially combine `custom_airflow_environment_mount_single` and `custom_airflow_environment_config` into a single helper.

Once these are combined, I think we could simplify the logic to say:

```
{{- if .Values.mountAllFromSecretName }}
envFrom:
{{- include "custom_airflow_environment_mount_all" . }}
env:
{{- include "standard_airflow_environment" . }}
{{- else }}
env:
{{- include "custom_airflow_environment_mount_individual" . }}
{{- include "standard_airflow_environment" . }}
{{- end -}}
{{- end -}}
```

This essentially makes it so that if you are using the new method, the old methods are completely ignored. If not, then it should behave exactly how it did before. What do you think?

templates/_helpers.yaml
kushalmalani: oh yeah. but this will redundant adding of env vars in the new mode. Should i take this out into a separate helper for the old way only?

schnie: Yea def is redundant, just thinking about backward compatibility. In my other comment I suggest combining both "old ways" (`Values.secret`, and `Values.env`) into a helper and adjusting the logic to allow for either the new way, or the old way, but not both at the same time. Essentially the new way will disable any of the old stuff.

templates/secrets/custom-airflow-secret.yaml
kushalmalani: We would need this if someone is going to use airflow charts in new mode, since with the flag enabled we tell pods to pick this secret file, which wouldnt exists and all pods fail with CreateContainerConfigError

PR #117, opened 2020-09-03
Fixing the init container config error with no secret file found

kushalmalani:
## Description

Fixing the init container config error with no secret file found

## PR Title

## 🎟 Issue(s)

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v1 is Helm 2
 apiVersion: v1
 name: airflow
-version: 0.15.5
+version: 0.15.6
 description: Helm chart to deploy the Astronomer Platform Airflow module
 icon: https://www.astronomer.io/static/airflowNewA.png
 keywords:
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -52,7 +52,20 @@
 {{- end }}
 
 {{/* User defined Airflow environment variables */}}
-{{- define "custom_airflow_environment" }}
+{{- define "airflow_environment_variables" }}
+{{- if .Values.mountAllFromSecretName }}
+envFrom:
+{{- include "custom_airflow_environment_mount_all" . }}
+env:
+{{- include "standard_airflow_environment" . }}
+{{- else }}
+env:
+{{- include "custom_airflow_environment_mount_individual" . }}
+{{- include "standard_airflow_environment" . }}
+{{- end -}}
+{{- end -}}
+
+{{- define "custom_airflow_environment_mount_individual" }}
   # Dynamically created environment variables
   {{- range $i, $config := .Values.env }}
   - name: {{ $config.name }}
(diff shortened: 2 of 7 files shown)


sjmiller609 (changes requested):
I think you need to account for the conditional https://github.com/astronomer/airflow-chart/blob/b37000e8737906ef6ba06fb4fa2ada6c0c8f7fff/templates/_helpers.yaml#L56

and also please complete the PR template.

kushalmalani:
> I think you need to account for the conditional
> 
> https://github.com/astronomer/airflow-chart/blob/b37000e8737906ef6ba06fb4fa2ada6c0c8f7fff/templates/_helpers.yaml#L56
> 
> and also please complete the PR template.

@sjmiller609 Init containers do not need custom environment variables. I am just adding the standard custom environment variables helper

sjmiller609:
I'm worried that this change might not address the problem. This feature is intended to make use of the all-at-once secret mount in order to avoid race conditions between the env var secret changing, and the deployment changing.

```
        envFrom:
        - secretRef:
            name: test-kube-env
```

Currently, this secret does not exist in the namespace in our test cluster
```
 $ k get secrets
NAME                                                       TYPE                                  DATA   AGE
astronomer-houston-jwt-signing-certificate                 Opaque                                1      19h
default-token-6xjrk                                        kubernetes.io/service-account-token   3      19h
sh.helm.release.v1.test-astrotest-2010.v1                  helm.sh/release.v1                    1      19h
test-astrotest-2010-airflow-metadata                       Opaque                                1      19h
test-astrotest-2010-airflow-result-backend                 Opaque                                1      19h
test-astrotest-2010-broker-url                             Opaque                                1      19h
test-astrotest-2010-cleanup-serviceaccount-token-kqp8t     kubernetes.io/service-account-token   3      19h
test-astrotest-2010-elasticsearch                          Opaque                                1      19h
test-astrotest-2010-fernet-key                             Opaque                                1      19h
test-astrotest-2010-pgbouncer-config                       Opaque                                2      19h
test-astrotest-2010-pgbouncer-stats                        Opaque                                1      19h
test-astrotest-2010-redis-password                         Opaque                                1      19h
test-astrotest-2010-registry                               kubernetes.io/dockerconfigjson        1      19h
test-astrotest-2010-scheduler-serviceaccount-token-jhwhk   kubernetes.io/service-account-token   3      19h
test-astrotest-2010-webserver-serviceaccount-token-4zphz   kubernetes.io/service-account-token   3      19h
test-astrotest-2010-worker-serviceaccount-token-znlk5      kubernetes.io/service-account-token   3      19h
```
and the airflow workloads are trying to consume it. I think this change does not address that issue.

kushalmalani:
The reason for that file not being there is because the pods never left the init container phase to app container phase. I am not 100% confident of the behavior on upgrade-deployments, what exactly the flow is. But in theory, houston will ask commander to create and empty secret file.

I cant test this locally, since there is no way for astronomer/astronomer to pick local airflow charts.

PR #120, opened 2020-09-25
Support for modifying celery worker deployment strategy

jpweber:
## Description

This PR modifies the worker template to allow passing a non-default deployment update strategy to worker deployments, in particular celery workers. The values have been set to allow 100% maxSurge and 50% maxUnavailable allowing new deployments of celery workers to launch a full set of replicas before the old set goes away. Allowing the new workers to pick up work as quickly as possible, rather than the current default which is 1 at a time.

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1410

## 🧪  Testing

helm chart unit test added. 

## 📋 Checklist

- [x] Helm chart unit test(s)

### DIFF
--- a/templates/workers/worker-deployment.yaml
+++ b/templates/workers/worker-deployment.yaml
@@ -26,6 +26,9 @@ spec:
       tier: airflow
       component: worker
       release: {{ .Release.Name }}
+  {{ if .Values.workers.updateStrategy }}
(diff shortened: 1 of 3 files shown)


schnie (approved):
Nice! LGTM

kaxil (approved):

PR #125, opened 2020-10-08
Add pod_template_file as that is the only way to use K8sExecutor in 2.0

dimberman:
## Description

## PR Title

## 🎟 Issue(s)

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -73,24 +73,27 @@ data:
     worker_container_repository = {{ .Values.images.airflow.repository }}
     worker_container_tag = {{ .Values.images.airflow.tag }}
     worker_container_image_pull_policy = {{ .Values.images.airflow.pullPolicy }}
-    worker_service_account_name = {{ .Release.Name }}-worker-serviceaccount
-    image_pull_secrets = {{ template "registry_secret" . }}
     dags_in_image = True
     delete_worker_pods = True
-
-    [kubernetes_secrets]
-    AIRFLOW__CORE__SQL_ALCHEMY_CONN = {{ printf "%s=connection" (include "airflow_metadata_secret" .) }}
-    AIRFLOW__CORE__FERNET_KEY = {{ printf "%s=fernet-key" (include "fernet_key_secret" .) }}
-{{- if .Values.elasticsearch.enabled }}
-    AIRFLOW__ELASTICSEARCH__ELASTICSEARCH_HOST = {{ printf "%s=connection" (include "elasticsearch_secret" .) }}
-{{- end }}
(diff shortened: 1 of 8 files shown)


dimberman:
@ernest-kr this should allow you to run k8sexecutor 2.0

danielhoherd (approved):

templates/configmap.yaml
```diff
@@ -73,8 +73,6 @@ data:
     worker_container_repository = {{ .Values.images.airflow.repository }}
     worker_container_tag = {{ .Values.images.airflow.tag }}
     worker_container_image_pull_policy = {{ .Values.images.airflow.pullPolicy }}
-    worker_service_account_name = {{ .Release.Name }}-worker-serviceaccount
-    image_pull_secrets = {{ template "registry_secret" . }}
```
schnie, line 77: @dimberman why'd we make these changes?

schnie:
@dimberman might need a rebase with main. Some of those changes should already be in there for 2.0.

templates/configmap.yaml
dimberman: @schnie these configs no longer matter when there is a pod_template_file, that said I could keep them in if we plan to keep using this chart for earlier airflow versions? (or maybe we might need to fork?)

schnie: We could either keep them or use the `semVerCompare` function here too. What do you think?

dimberman: @schnie that works I'll set that up now.

templates/configmap.yaml
```diff
@@ -73,25 +73,28 @@ data:
...
-    AIRFLOW__CORE__FERNET_KEY = {{ printf "%s=fernet-key" (include "fernet_key_secret" .) }}
-{{- if .Values.elasticsearch.enabled }}
-    AIRFLOW__ELASTICSEARCH__ELASTICSEARCH_HOST = {{ printf "%s=connection" (include "elasticsearch_secret" .) }}
-{{- end }}
-
+    {{- if semverCompare ">=1.10.12" .Values.airflowVersion }}
+    pod_template_file = {{ include "airflow_pod_template_file" . }}/pod_template_file.yaml
+    {{- else }}
+    worker_service_account_name = {{ .Release.Name }}-worker-serviceaccount
+    image_pull_secrets = {{ template "registry_secret" . }}
```
schnie, line 82:
```suggestion
    image_pull_secrets = {{ template "registry_secret" . }}

```

PR #127, opened 2020-10-15
Add support for celery worker volume attachment 

ianstanton:
## Description

- This allows for attaching a persistent volume to celery workers.
- A user can make use of this with a configuration similar to the following:
```
executor: CeleryExecutor
workers:
  additionalVolume:
    enabled: true
    storageClassName: efs-sc
    mountPath: /usr/local/airflow/efs
    volumePlugin:
      csi:
        driver: efs.csi.aws.com
        volumeHandle: fs-5c044359
```

## 🎟 Issue(s)

Resolves astronomer/issues#1954

## 🧪  Testing
- Attempted to add helm unit tests and found with @danielhoherd that there are some bugs in the framework
- Leaving tests out until the following is resolved https://github.com/astronomer/helm-unittest/issues/2


schnie (approved):
LGTM

schnie:
@ianstanton looks great, maybe just add a test or two for these? There should be a file for celery workers under `tests` directory already that you can add to.

danielhoherd:
I worked with Ian on adding tests here and we found that no matter what we did, the tests passed. I've opened a ticket to look into that issue: astronomer/helm-unittest#2

PR #134, opened 2020-10-22
Add webserver_config.py configuration value

serialx:
## Description

This PR adds webserver_config.py configuration to the helm chart values. Normally authentication settings require custom webserver_config.py setting. Which is currently not supported by Astronomer airflow chart.

## PR Title

Add webserver_config.py configuration value

## 🧪  Testing

This should be backwards compatible.

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -101,6 +101,10 @@ data:
     jwt_audience = {{ printf "%s/%s" ( include "deployments_subdomain" . ) .Release.Name }}
 {{- end }}
 
+{{- if .Values.webserver.webserverConfig }}
+  webserver_config.py: |
+    {{ .Values.webserver.webserverConfig | nindent 4 }}
+{{- else }}
   webserver_config.py: |
     import os
     from airflow import configuration as conf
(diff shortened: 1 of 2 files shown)


danielhoherd:
@serialx please add [tests](https://github.com/astronomer/airflow-chart/tree/master/tests).

templates/configmap.yaml
```diff
@@ -101,6 +101,10 @@ data:
     jwt_audience = {{ printf "%s/%s" ( include "deployments_subdomain" . ) .Release.Name }}
 {{- end }}
 
+{{- if .Values.webserver.webserverConfig }}
+  webserver_config.py: |
+    {{ .Values.webserver.webserverConfig | nindent 4 }}
+{{- else }}
   webserver_config.py: |
```
ashb, line 108:
```suggestion
  webserver_config.py: |
{{- if .Values.webserver.webserverConfig }}
    {{ .Values.webserver.webserverConfig | nindent 4 }}
{{- else }}
```

ashb (changes requested):
And tests required as Daniel says :)

PR #136, opened 2020-10-22
Add logging volume to pod_template_file

dimberman:
## Description

## PR Title

## 🎟 Issue(s)

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/files/pod-template-file.yaml
+++ b/files/pod-template-file.yaml
@@ -1,6 +1,15 @@
 ---
+# Are we using a local/sequenial executor?
+{{- $local := or (eq .Values.executor "LocalExecutor") (eq .Values.executor "SequentialExecutor") }}
+# Are we using the kubernetes executor?
+{{- $kube := eq .Values.executor "KubernetesExecutor" }}
+# Is persistence enabled on the _workers_?
+# This is important because in $local mode, the scheduler assumes the role of the worker
+{{- $persistence := .Values.workers.persistence.enabled }}
 # If we're using a StatefulSet
 {{- $stateful := and $local $persistence }}
+# If we're using elasticsearch logging
+{{- $elasticsearch := .Values.elasticsearch.enabled }}
 
 apiVersion: v1
 kind: Pod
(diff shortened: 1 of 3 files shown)


files/pod-template-file.yaml
```diff
@@ -1,4 +1,16 @@
 ---
+# Are we using a local/sequenial executor?
```
kaxil, line 2:
```suggestion
# Are we using a local/sequential executor?
```

files/pod-template-file.yaml
```diff
@@ -48,3 +60,19 @@ spec:
   - configMap:
       name: {{ include "airflow_config" . }}
     name: airflow-local-settings
+{{- if not $stateful }}
+  - name: logs
```
kaxil, line 64: does this need to be `airflow-logs`?

files/pod-template-file.yaml
```diff
@@ -48,3 +60,19 @@ spec:
...
+    emptyDir: {}
+{{- else }}
+  volumeClaimTemplates:
+    - metadata:
+      name: logs
```
kaxil, line 69: same here

templates/configmap.yaml
```diff
@@ -75,6 +75,7 @@ data:
     worker_container_image_pull_policy = {{ .Values.images.airflow.pullPolicy }}
     dags_in_image = True
     delete_worker_pods = True
+    delete_on_success = False
```
kaxil, line 78: is this needed, or just leftover from your debugging?

dimberman: @kaxil I think in general this is the best setting for debugging (basically only keeps pods up if they fail), though I'm willing to hear otherwise if that's not ideal for cloud

kaxil: yea I fear for cloud customers, as they can't delete the pods, this will increase their usage.

cc @schnie

kaxil, line 78:
```suggestion
```

kaxil: I have removed it from this PR to unblock merging of this PR, feel to create a new PR to add that

kaxil (approved):

PR #138, opened 2020-10-28
Reduce CI job complexity. Use quay.io/astronomer/ci-helm-release

danielhoherd:
## Description

Reduce the complexity of the helm release process by using a new docker image I've created for doing that task. Also clean up a bunch of shell script code, fix shellcheck complaints, and add pre-commit config.
## 🧪  Testing

These are CI changes only, so there's no risks for our customers, just the potential for CI to misbehave.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
 
   build-artifact:
     docker:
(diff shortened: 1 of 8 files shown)


jpweber (approved):
great improvements.

PR #145, opened 2020-11-04
Add namespace in the port-forward command

msumit:
Could not test by running the chart locally, as it was failing with the following error `Error: found in Chart.yaml, but missing in charts/ directory: postgresql`


ashb (approved):

ashb:
Thanks Sumit!

PR #151, opened 2020-11-09
Adding support for Openshift Security Context Constraints

sjmiller609:
## Description

Peer programmed with @ianstanton 

Adding support for Openshift Security Context Constraints

## 🎟 Issue(s)

Part of https://github.com/astronomer/issues/issues/2039

## 🧪  Testing

This configuration is default disabled. For the enabled case, a unit test has been added.

## 📋 Checklist

- [x] Helm chart unit test(s)

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v1 is Helm 2
 apiVersion: v1
 name: airflow
-version: 0.18.1
+version: 0.18.0
 description: Helm chart to deploy the Astronomer Platform Airflow module
 icon: https://www.astronomer.io/static/airflowNewA.png
 keywords:
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v1 is Helm 2
 apiVersion: v1
 name: airflow
-version: 0.18.0
+version: 0.18.1
 description: Helm chart to deploy the Astronomer Platform Airflow module
 icon: https://www.astronomer.io/static/airflowNewA.png
 keywords:
--- a/templates/airflow-scc-anyuid.yaml
+++ b/templates/airflow-scc-anyuid.yaml
@@ -38,6 +38,7 @@ allowHostIPC: false
 allowHostNetwork: false
 allowHostPID: false
 allowHostPorts: false
+allowPrivilegeEscalation: false
 allowPrivilegedContainer: false
 allowedCapabilities: null
 {{- end }}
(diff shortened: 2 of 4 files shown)


templates/airflow-scc-anyuid.yaml
```diff
@@ -0,0 +1,44 @@
...
+fsGroup:
+  type: RunAsAny
+groups:
+- system:cluster-admins
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: true
```
schnie, line 41: Does this need to be `true` for something to work?

ianstanton: @sjmiller609 and I confirmed we do not need `allowPrivilegeEscalation: true`. We'll remove this.

templates/airflow-scc-anyuid.yaml
```diff
@@ -0,0 +1,44 @@
...
+- downwardAPI
+- emptyDir
+- persistentVolumeClaim
+- projected
+- secret
+defaultAddCapabilities: null
+fsGroup:
+  type: RunAsAny
+groups:
+- system:cluster-admins
```
schnie, line 35: What does this `groups` section do?

ianstanton: We likely don't need this `groups` section as we have a list of users above. Some info on groups (a collection of users) https://docs.openshift.com/container-platform/3.9/admin_guide/manage_users.html#admin-guide-managing-users-creating-groups

templates/airflow-scc-anyuid.yaml
ianstanton: We found this needs to be true (at least with Debian image). Without this, we receive the following error:

```
▶ oc logs airflow-scheduler-7d46df6bb6-28lst -n airflow -c wait-for-airflow-migrations
error: failed switching to "astro": operation not permitted
```

templates/airflow-scc-anyuid.yaml
ianstanton: Found that this is not necessary. Removed this section.

ianstanton (approved):

Chart.yaml
```diff
@@ -1,7 +1,7 @@
 # apiVersion v1 is Helm 2
 apiVersion: v1
 name: airflow
-version: 0.18.0
+version: 0.18.1
```
ashb, line 4: I don't think 0.18.0 has actually been released yet (but I'm not 100% sure) -- we're still testing changes for Airflow 2.0 here.

sjmiller609, line 4:
```suggestion
version: 0.18.0
```

PR #153, opened 2020-11-11
Fix spacing for affinity

dimberman:
The affinity, nodeselector, and tolerations settings were
indented too far. This was causing failures when platform
affinities were added to the pod

## Description

## PR Title

## 🎟 Issue(s)

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/files/pod-template-file.yaml
+++ b/files/pod-template-file.yaml
@@ -51,11 +51,11 @@ spec:
     runAsUser: 50000
   {{- end }}
   nodeSelector:
-    {{ toYaml .Values.nodeSelector | indent 8 }}
+    {{ toYaml .Values.nodeSelector | indent 4 }}
   affinity:
-    {{ toYaml .Values.affinity | indent 8 }}
+    {{ toYaml .Values.affinity | indent 4 }}
   tolerations:
-    {{ toYaml .Values.tolerations | indent 8 }}
+    {{ toYaml .Values.tolerations | indent 4 }}
   serviceAccountName: '{{ .Release.Name }}-worker-serviceaccount'
   volumes:
   - configMap:
--- a/files/pod-template-file.yaml
+++ b/files/pod-template-file.yaml
@@ -47,15 +47,12 @@ spec:
   securityContext:
   {{- if contains "alpine"  .Values.defaultAirflowTag }}
     runAsUser: 100
-  {{- else }}g
+  {{- else }}
     runAsUser: 50000
   {{- end }}
-  nodeSelector:
-{{ toYaml .Values.nodeSelector | indent 4 }}
-  affinity:
-{{ toYaml .Values.affinity | indent 4 }}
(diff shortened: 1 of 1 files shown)


files/pod-template-file.yaml
```diff
@@ -47,15 +47,15 @@ spec:
   securityContext:
   {{- if contains "alpine"  .Values.defaultAirflowTag }}
     runAsUser: 100
-  {{- else }}
+  {{- else }}g
```
jpweber, line 50: Looks like a stray `g` here.

files/pod-template-file.yaml
```diff
@@ -47,15 +47,15 @@ spec:
...
     runAsUser: 50000
   {{- end }}
   nodeSelector:
-    {{ toYaml .Values.nodeSelector | indent 8 }}
+{{ toYaml .Values.nodeSelector | indent 4 }}
```
jpweber, line 54: Something that will make this easier is the use of `nindent` makes a new line then indents the value, and does not depend at all on the spacing where `{{ toYaml` starts. For example.
```
nodeSelector: {{ toYaml .Values.nodeSelector | nindent 4 }}
```
Also arguably easier to read.

jpweber (changes requested):
Inline comments, plus helm chart unit-tests are failing. When it fails with tons of errors like that, it means something is likely wrong higher up in the fail that just causes the whole thing to bonk out.

dimberman:
Thanks @jpweber fixed

files/pod-template-file.yaml
```diff
@@ -47,15 +47,12 @@ spec:
...
+  {{- else }}
     runAsUser: 50000
   {{- end }}
-  nodeSelector:
-{{ toYaml .Values.nodeSelector | indent 4 }}
-  affinity:
-{{ toYaml .Values.affinity | indent 4 }}
-  tolerations:
-{{ toYaml .Values.tolerations | indent 4 }}
+  nodeSelector: {{ toYaml .Values.nodeSelector | indent 4 }}
```
jpweber, line 53: `indent` needs to be `nindent`  notice the `n` at the beginning. you need the `n` to start a newline when it puts your values in.

jpweber (changes requested):

jpweber (approved):

ernest-kr:
I guess the helm chart would be automatically pushed the internal and main repos right
@jpweber

jpweber:
@ernest-kr a new helm chart is built. according to the CI the chart version is `0.18.0-build7256` But nothing is automatically deployed anywhere. That is a manual step.

ernest-kr:
ok I will ask the platform team to use this chart version

ashb:
@dimberman Could you follow up this PR to add tests please.

PR #154, opened 2020-11-11
Update chart to use both 2.0 and 1.10 name of config options

ashb:
Without this we get lots of deprecation warnings in the logs

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -19,8 +19,16 @@ data:
   airflow.cfg: |
     [core]
     load_examples = False
-    colored_console_log = False
     executor = {{ .Values.executor }}
+
+{{/* Support old/pre-2.0 config names */}}
+    colored_console_log = False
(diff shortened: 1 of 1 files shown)


kaxil (approved):

templates/configmap.yaml
```diff
@@ -25,6 +25,12 @@ data:
     remote_logging = True
 {{- end }}
 
+    [logging]
+    colored_console_log = False
+{{- if .Values.elasticsearch.enabled }}
```
kaxil, line 30: We should do this for statsd too:

https://github.com/apache/airflow/blob/0d51a12e26cdc5b291e0b9855852576ad8e610f6/UPDATING.md#metrics-configuration-has-been-moved-to-new-section

ashb: Oh yes, done.

kaxil (approved):

PR #157, opened 2020-11-17
Prevent Airflow Webserver Reboot for Airflow 2.0+

andriisoldatenko:
## Description

Prevent Airflow Webserver Reboot for Airflow 2.0+

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/2123

### DIFF
--- a/templates/webserver/webserver-deployment.yaml
+++ b/templates/webserver/webserver-deployment.yaml
@@ -65,7 +65,7 @@ spec:
           {{- include "standard_airflow_environment" . | indent 10 }}
       containers:
         - name: webserver
-          {{- if .Values.webserverReboot }}
+          {{- if .Values.webserver.useDefaultAirflowImage }}
           image: {{ template "airflow_image" . }}
           {{- else }}
           image: {{ template "default_airflow_image" . }}
--- a/templates/webserver/webserver-deployment.yaml
+++ b/templates/webserver/webserver-deployment.yaml
@@ -65,10 +65,10 @@ spec:
           {{- include "standard_airflow_environment" . | indent 10 }}
       containers:
         - name: webserver
-          {{- if semverCompare ">=2.0.0" .Values.airflowVersion }}
-          image: {{ template "default_airflow_image" . }}
-          {{- else }}
+          {{- if .Values.webserver.reboot }}
           image: {{ template "airflow_image" . }}
+          {{- else }}
+          image: {{ template "default_airflow_image" . }}
           {{- end }}
           imagePullPolicy: {{ .Values.images.airflow.pullPolicy }}
           args: ["airflow", "webserver"]
--- a/templates/webserver/webserver-deployment.yaml
+++ b/templates/webserver/webserver-deployment.yaml
@@ -65,7 +65,11 @@ spec:
           {{- include "standard_airflow_environment" . | indent 10 }}
       containers:
         - name: webserver
+          {{- if semverCompare ">=1.10.12" .Values.airflowVersion }}
+          image: {{ template "default_airflow_image" . }}
+          {{- else }}
           image: {{ template "airflow_image" . }}
+          {{- end }}
           imagePullPolicy: {{ .Values.images.airflow.pullPolicy }}
           args: ["airflow", "webserver"]
           resources:
--- a/templates/webserver/webserver-deployment.yaml
+++ b/templates/webserver/webserver-deployment.yaml
@@ -65,7 +65,7 @@ spec:
           {{- include "standard_airflow_environment" . | indent 10 }}
       containers:
         - name: webserver
-          {{- if .Values.webserver.reboot }}
+          {{- if .Values.webserverReboot }}
           image: {{ template "airflow_image" . }}
           {{- else }}
           image: {{ template "default_airflow_image" . }}
--- a/templates/webserver/webserver-deployment.yaml
+++ b/templates/webserver/webserver-deployment.yaml
@@ -65,7 +65,7 @@ spec:
           {{- include "standard_airflow_environment" . | indent 10 }}
       containers:
         - name: webserver
-          {{- if semverCompare ">=1.10.12" .Values.airflowVersion }}
+          {{- if semverCompare ">=2.0.0" .Values.airflowVersion }}
           image: {{ template "default_airflow_image" . }}
           {{- else }}
           image: {{ template "airflow_image" . }}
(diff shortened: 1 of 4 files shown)


schnie:
@andriisoldatenko looks good!

@astrorfox did we decide we want to prevent webserver reboots for 2.0+? I think we talked about this and decided to push it from 1.10.12 to 2.0, even though serialization is enabled by default in 1.10.12 already.

schnie (approved):

schnie:
We moved this to 2.0 because the option to override this flag is officially removed. 1.10.12 can still override to disable which would create an issue.

templates/webserver/webserver-deployment.yaml
```diff
@@ -65,7 +65,11 @@ spec:
           {{- include "standard_airflow_environment" . | indent 10 }}
       containers:
         - name: webserver
+          {{- if semverCompare ">=2.0.0" .Values.airflowVersion }}
+          image: {{ template "default_airflow_image" . }}
```
ashb, line 69: This entirely removes the ability to use plugins in the webserver.

I feel this needs to be opt-in per deployment somehow.

cc @schnie

ashb: Is the goal to stop rebooting, or to stop down time of webserver on deploy?

If the later we could change the replacement policy on the deployment

andriisoldatenko: good catch, do you mean change pullPolicy for image?

schnie: The ultimate goal is to prevent seeing the rocketship after they do a code-push. People don't like the webserver downtime. Forgot about plugins/blueprints/custom views though. Dang, that makes it tough.

astrorfox: Could we change the default to not restart and allow the user to override when needed?

schnie: There may be a better UX, but off the top of my head maybe it could be a checkbox on the webserver settings to "Allow Plugins", or something. Under the hood we use that flag to determine if we use `default_airflow_image` or `airflow_image`. 

The caveat of "Allowing Plugins" would be that we keep the current behavior and reboot the webserver.

I _feel_ like a small amount of customers user plugins, so they'd be the exception here, but I really don't know for sure. @paolaperaza @ryw @vparekh94 @petedejoy any idea on this?

paolaperaza: (FYI - summarized the convo/questions that have come up on this in the GH issue itself: https://github.com/astronomer/issues/issues/2123#issuecomment-730033302)

tests/webserver_webserver-deployment_test.yaml
```diff
@@ -7,3 +7,21 @@ tests:
...
+    set:
+      airflowVersion: 1.10.11
+      images:
+        airflow:
+          tag: 1.10.11-custom
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].image
+          value: quay.io/astronomer/ap-airflow:1.10.11-custom
+  - it: should get correct airflow container image for 1.10.12
```
vishwas-astro, line 20: Should description be 2.0.0 instead of 1.10.12?

andriisoldatenko, line 20:
```suggestion
  - it: should get correct airflow container image for 2.0.0
```

values.yaml
```diff
@@ -263,6 +263,9 @@ scheduler:
   # Number of schedulers to run
   replicas: 1
 
+# Prevent airflow webserver reboot only for Airflow 2.0.0
+webserverReboot: true
```
schnie, line 267: @andriisoldatenko what do you think about nesting the new flag under the existing `webserver` section?

Thinking maybe we could name it something like `webserver.useDefaultAirflowImage` to tie the underlying intention of the flag back to the actual underlying helper name. Just so its a little clearer as to what's actually changing under the hood. What do you think?

andriisoldatenko:
@schnie PTAL

values.yaml
andriisoldatenko: @schnie done.

schnie:
LGTM

tests/webserver_webserver-deployment_test.yaml
```diff
@@ -7,3 +7,25 @@ tests:
...
+        airflow:
+          tag: 1.10.11-custom
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].image
+          value: quay.io/astronomer/ap-airflow:1.10.11-custom
+  - it: should get correct airflow container image for 2.0.0
+    set:
+      webserver:
+        useDefaultAirflowImage: true
```
vishwas-astro, line 25: Could there be a test case where you dont specify any value for useDefaultAirflowImage at all and it should assume the default of true, and hence there will be a restart? 

Not a blocker for merge. Just a curious question.

andriisoldatenko:
@ashb PTAL

tests/webserver_webserver-deployment_test.yaml
andriisoldatenko: default value is true https://github.com/astronomer/airflow-chart/pull/157/files#diff-8377b3e3740a3fcd9f682e5fb55425f2fdbece1791854b9e5013e7f1a5e60e7eR284

values.yaml
```diff
@@ -280,6 +280,9 @@ webserver:
   # Number of webservers
   replicas: 1
 
+  # Prevent airflow webserver reboot only for Airflow 2.0.0
+  useDefaultAirflowImage: true
```
kaxil, line 284:
```suggestion
  # Prevent Airflow webserver reboot only for Airflow 2.0.0
  # AC image with matching Airflow version will be used
  # to run the Webserver. Since Airflow >= 2.0 uses DAG Serialization
  # User image containing DAGs is not required to run Webserver.
  useDefaultAirflowImage: true
```

Can you add this info to https://github.com/astronomer/airflow-chart/blob/master/README.md#parameters too please

kaxil: We should also add docs to https://www.astronomer.io/docs/cloud/stable/deploy/deploy-cli -- not as a scope of this PR but maybe as and additional Github Issue that can be tackled separately.

cc @paolaperaza

schnie, line 284:
```suggestion
  # Prevent airflow webserver reboot only for Airflow 2.0.0
  useDefaultAirflowImage: false
```

Setting this to `false` by default until we have 2.0 as the default image.

schnie:
I set the default value to `false` to be on the safe side since we do not have 2.0 as the default image in the chart. Houston will override to `true` when it needs to.

I think this is good to go now.

andriisoldatenko:
@kaxil @vishwas-astro PTAL

vishwas-astro (approved):

kaxil (approved):

PR #159, opened 2020-11-19
Change default uid/gid to values for Debian image

ashb:
This matches the current default airflow tag used in the chart.

Part of astronomer/issues#2100 -- there will be a companion change in Houston.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -3,8 +3,8 @@
 # Declare variables to be passed into your templates.
 
 # User and group of airflow user
-uid: 100
-gid: 101
+uid: 50000
+gid: 50000
 
 # Select certain nodes for airflow pods.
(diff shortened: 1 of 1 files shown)


andriisoldatenko (approved):

kaxil (approved):

ashb:
We want this in the next release, but probably need to hold off merging this until https://github.com/astronomer/houston-api/pull/581 is done, right @andriisoldatenko ?

andriisoldatenko:
> We want this in the next release, but probably need to hold off merging this until [astronomer/houston-api#581](https://github.com/astronomer/houston-api/pull/581) is done, right @andriisoldatenko ?

yes, we need to double check generateHelm values output. So let's wait on houston-api PR.

PR #163, opened 2020-11-25
Include custom environment variables in pod-template-file

schnie:


jpweber (approved):

files/pod-template-file.yaml
```diff
@@ -26,11 +26,19 @@ spec:
     - args: []
       name: "base"
       command: []
+      {{/* Environment section duplicated from _helpers.yaml file */}}
+      {{- if .Values.mountAllFromSecretName }}
```
danielhoherd, line 30: This variable, `mountAllFromSecretName`, is vague. As part of this change can we add a comment in `values.yaml` documenting what this variable is for?

dimberman (approved):

kushalmalani (approved):
LGTM

files/pod-template-file.yaml
schnie: done

PR #166, opened 2020-12-01
Allow webserver to read pod logs

dimberman:
Allows webserver to read pod logs for tasks launched by k8sexecutor
without needing an external log store. Note: logs will only be
accessible while the pod is running (or after if you don't delete it)

## Description

## PR Title

## 🎟 Issue(s)

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/templates/rbac/pod-log-reader-role.yaml
+++ b/templates/rbac/pod-log-reader-role.yaml
@@ -1,0 +1,56 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+################################
+## Airflow Pod Reader Role
+#################################
+{{- if and .Values.rbacEnabled .Values.webserver.allowPodLogReading }}
+{{- if .Values.multiNamespaceMode }}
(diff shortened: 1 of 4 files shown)


templates/rbac/pod-log-reader-role.yaml
```diff
@@ -0,0 +1,56 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+################################
+## Airflow Pod Reader Role
+#################################
+{{- if and .Values.rbacEnabled .Values.webserver.allowPodLogReading }}
```
ashb, line 21: Do we need a separate config here, or could we re-use the `.Values.allowPodLaunching` here?

values.yaml
```diff
@@ -267,6 +267,7 @@ scheduler:
 
 # Airflow webserver settings
 webserver:
+  allowPodLogReading: true
```
ashb, line 270: Do we want this false by default in our chart maybe?

dimberman: @ashb it shouldn't matter since we have external log storage, but glad to do so if you feel it's more secure.

ashb (approved):

vishwas-astro:
@ashb @dimberman I am assuming this fix does not need to make it into the 0.23.x platform release. Is that a valid assumption?

cc @paolaperaza @vikramkoka

vikramkoka:
> @ashb @dimberman I am assuming this fix does not need to make it into the 0.23.x platform release. Is that a valid assumption?
> 
> cc @paolaperaza @vikramkoka

@dimberman I did not realize earlier today that this was a fix into the helm chart. Since the OSS Helm chart is not being released as part of 2.0, I am completely fine deferring this to avoid unnecessary code changes just before release.

ashb:
This doesn't affect users of our platform - only people using the chart outside of it.

So no, not needed in 0.23.1

guidok91:
Hey guys,

Any estimation of when this will be released?

We are facing [an issue](https://github.com/apache/airflow/issues/11696) when reading logs from the UI (introduced with [this Airflow PR](https://github.com/apache/airflow/pull/8626/files)).

As far as I'm aware, the issue would be fixed with this PR.

PR #174, opened 2020-12-11
Remove redundant flower_image

andriisoldatenko:
Resolves astronomer/issues#2368


schnie (commented):
LGTM, should we remove this block too while we're at it? https://github.com/astronomer/airflow-chart/blob/master/values.yaml#L91-L94

andriisoldatenko:
> LGTM, should we remove this block too while we're at it? https://github.com/astronomer/airflow-chart/blob/master/values.yaml#L91-L94

done.

ashb (approved):

kaxil (approved):

Issue #176, opened 2020-12-19
readme keda install instructions produce ScaledObject error


dstandish:
I tried following the instructions for keda install (installing locally in k3d)

I got this error:

```
❯ helm install airflow \                                                                                                                                                                             <aws:edm-dstandish>
    --set executor=CeleryExecutor \
    --set workers.keda.enabled=true \
    --set workers.persistence.enabled=false \
    --namespace airflow \
    astronomer/airflow
Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ScaledObject" in version "keda.k8s.io/v1alpha1"
```

PR #177, opened 2020-12-19
fix keda version in readme

dstandish:
This chart expects keda version 1.5.0.  Here i update readme to reference it.

closes https://github.com/astronomer/airflow-chart/issues/176


dstandish:
closes https://github.com/astronomer/airflow-chart/issues/176

danielhoherd (approved):

PR #179, opened 2020-12-22
Switch to quay.io on release-15

danielhoherd:
Switch to using Quay.io in chart version 0.15, which is what LTS users use.

Related to astronomer/issues#1355


adam2k (approved):
🥳

andriisoldatenko:
LGTM, just reminder to bump it here too
https://github.com/astronomer/astronomer/blob/release-0.16/charts/astronomer/values.yaml#L8

andriisoldatenko (approved):

vishwas-astro (approved):

PR #184, opened 2021-02-01
add ability to add more settings to pgbouncer config secret

tommyk:
# overview

we are looking to be able to add more settings to the pgbouncer config file.  this is a less targeted, more open, approach to allowing a user to add configuration to a few places in the pgbouncer configuration file.

i wanted to open this up to get the conversation going early about this approach and just get any feedback people had.  i will be adding tests here in a bit.

let me know your thoughts!

### DIFF
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -173,8 +173,8 @@ env:
 
 {{ define "pgbouncer_config" }}
 [databases]
-{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }}
-{{ .Release.Name }}-result-backend = host={{ .Values.data.resultBackendConnection.host }} dbname={{ .Values.data.resultBackendConnection.db }} port={{ .Values.data.resultBackendConnection.port }} pool_size={{ .Values.pgbouncer.resultBackendPoolSize }}
+{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }} {{ .Values.pgbouncer.extraIniDatabaseMetatdata }}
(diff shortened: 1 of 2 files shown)


tommyk:
if you want to see the output easily, here is a quick one liner to produce it
```bash
 helm package . --dependency-update && helm install \                                               
  --dry-run \
  --set pgbouncer.enabled=true \
  --set pgbouncer.metadataPoolSize=25 \
  --set pgbouncer.resultBackendPoolSize=25 \
  --set data.metadataConnection.host="airflow-postgresql.airflow.svc.cluster.local" \
  --set data.resultBackendConnection.host="airflow-postgresql.airflow.svc.cluster.local" \
  test airflow-0.18.5.tgz | grep "pgbouncer.ini:" | sed -E 's/pgbouncer.ini://' | base64 -D
```

ashb (approved):
That diff is hard to read :) Assuming you've tested this LGTM

templates/_helpers.yaml
```diff
@@ -173,8 +173,8 @@ env:
 
 {{ define "pgbouncer_config" }}
 [databases]
-{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }}
-{{ .Release.Name }}-result-backend = host={{ .Values.data.resultBackendConnection.host }} dbname={{ .Values.data.resultBackendConnection.db }} port={{ .Values.data.resultBackendConnection.port }} pool_size={{ .Values.pgbouncer.resultBackendPoolSize }}
+{{ .Release.Name }}-metadata = host={{ .Values.data.metadataConnection.host }} dbname={{ .Values.data.metadataConnection.db }} port={{ .Values.data.metadataConnection.port }} pool_size={{ .Values.pgbouncer.metadataPoolSize }} {{ .Values.pgbouncer.extraIniDatabaseMetatdata }}
+{{ .Release.Name }}-result-backend = host={{ .Values.data.resultBackendConnection.host }} dbname={{ .Values.data.resultBackendConnection.db }} port={{ .Values.data.resultBackendConnection.port }} pool_size={{ .Values.pgbouncer.resultBackendPoolSize }} {{ .Values.pgbouncer.extraIniDatabaseResultBackend }}
```
ashb, line 177: I do wonder if this works be clearer (and as flexible) by making. Values.data.resultBackendConnection an dict where the keys match pgbouncer settings, and then we just iterate over them?

Example values:
```yaml
resultBackendConnection:
  dbname: foo
  pool_size: 10
```

Though switching to this is a bit more complex on the generation side I guess

PR #187, opened 2021-02-12
Add fsgroup for IAM credentials files

rbankston:
## Description

Add fsGroup to the Kubernetes Pod Template definition. The current pod operator template overrides any template from the airflow image and if an IAM credential exists it won't be added with the proper ownership causing issues with it being accessed by the pod

## 🎟 Issue(s)

Resolves astronomer/issues#2587

## 📋 Checklist

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v1 is Helm 2
 apiVersion: v1
 name: airflow
-version: 0.18.5
+version: 0.18.6
 description: Helm chart to deploy the Astronomer Platform Airflow module
 icon: https://www.astronomer.io/static/airflowNewA.png
 keywords:
--- a/files/pod-template-file.yaml
+++ b/files/pod-template-file.yaml
@@ -62,6 +62,7 @@ spec:
   restartPolicy: Never
   securityContext:
     runAsUser: {{ .Values.uid }}
+    fsGroup: {{ .Values.gid }}
   nodeSelector: {{ toYaml .Values.nodeSelector | nindent 4 }}
   affinity: {{ toYaml .Values.affinity | nindent 4 }}
   tolerations: {{ toYaml .Values.tolerations | nindent 4 }}


rbankston:
@jpweber @dimberman @andriisoldatenko can you folks take a look at this for me?

ianstanton:
Taking a look

jpweber (approved):

ianstanton (changes requested):
Thanks @rbankston! Can we add `fsGroup` as a top level value? This will give the user the option to supply their own value. We can default to `.Values.gid` and use `.Values.fsGroup` if it exists. We can also add this option to the Celery worker deployment. What do you think? cc @ben-astro

rbankston:
@ianstanton https://github.com/astronomer/airflow-chart/blob/83856c33ef059ba44d022fa6b5c737d1db0eafcd/templates/workers/worker-deployment.yaml#L65 and https://github.com/astronomer/airflow-chart/blob/5934a9ce1c19f44f5067fba205b119d43df6030b/templates/scheduler/scheduler-deployment.yaml#L75 already use that value as the fsGroup by default

ianstanton:
@rbankston I guess I was thinking of adding `fsGroup: ~` to `values.yaml` and then using something like `fsGroup: {{ (default .Values.gid .Values.fsGroup }}` in case the user wanted to supply their own value for `fsGroup`. 

Do you think there will ever be a case where a user would want to supply their own value?

rbankston:
@ianstanton not 100% sure how Astronomer Certified folks would use that but don't believe today that Astronomer Enterprise folks can change the user of the pod easily.

ianstanton:
@rbankston Okay! we can leave it like this and revisit if the topic comes up again.

ianstanton (approved):

kaxil:
@danielhoherd the test is failing with:

```
bin/package-and-release-internal: line 16: GCP_TOKEN: unbound variable
```

Does it ring any bells?

rbankston:
Closing this pull request because of the way it was created. Same PR was merged in https://github.com/astronomer/airflow-chart/pull/190

PR #191, opened 2021-03-15
Add patch ability to pod-launcher-role

dimberman:
In order for the KubernetesExecutor to pick up tasks from dead
schedulers, it needs to have the ability to patch existing pods s.t. the
watcher thread can read the pod statuses

## Description

## PR Title

## 🎟 Issue(s)

Resolves astronomer/issues#2682

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/templates/rbac/pod-launcher-role.yaml
+++ b/templates/rbac/pod-launcher-role.yaml
@@ -23,6 +23,7 @@ rules:
       - "create"
       - "list"
       - "get"
+      - "patch"
       - "watch"
       - "delete"
   - apiGroups:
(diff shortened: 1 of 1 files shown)


ianstanton (approved):
@dimberman Simple enough, LGTM. The 'list events' piece is necessary for this as well?

jedcunningham (approved):
LGTM as well, but same question about events (as the commit message specifically mentions patch).

dimberman:
@ianstanton oh I just realized the list events is for this bug https://github.com/apache/airflow/commit/90ab60bba877c65cb93871b97db13a179820d28b

So it's still necessary, but not exactly for here. Should I split that into a different PR or just mention it in the description?

danielhoherd (approved):

PR #202, opened 2021-03-29
Update redis-statefulset.yaml to run as non-root

povode:
## Description

It's best practice to run the service as non-root. Some cluster enforce this as a matter of policy.
UID 999 is already present in the docker image.

## PR Title

Add RunAsUser to the Redis statefulset.

## 🎟 Issue(s)

No issue logged for this, I can log it if needed.

## 🧪  Testing

Pods from the generated Statefulset is working as non-root user.

### DIFF
--- a/templates/redis/redis-statefulset.yaml
+++ b/templates/redis/redis-statefulset.yaml
@@ -36,6 +36,8 @@ spec:
         cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
         {{- end }}
     spec:
+      securityContext:
+        RunAsUser: 999
       nodeSelector:
 {{ toYaml .Values.nodeSelector | indent 8 }}
       affinity:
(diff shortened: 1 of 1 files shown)


povode:
Error from CircleCi "bin/package-and-release-internal: line 16: GCP_TOKEN: unbound variable" doesn't seem to be related to changes I've made can some admin take a look?

schnie (approved):

templates/redis/redis-statefulset.yaml
```diff
@@ -36,6 +36,8 @@ spec:
         cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
         {{- end }}
     spec:
+      securityContext:
+        RunAsUser: 999
```
kaxil, line 40: This should be `runAsUser` right?

Please add tests too

jedcunningham: Also, while 999 is right for the official image, we may want to expose it as a parameter in case someone wants to use another redis image.

That said, the official image uses gosu to drop root anyways 🤷‍♂️

PR #204, opened 2021-04-01
Add pgbouncer.networkPolicies.enabled flag

danielhoherd:
* Add pgbouncer.networkPolicies.enabled flag
* Document pod-template-file behavior
* Auto-format via vscode 'prettier' plugin

This should solve https://github.com/astronomer/issues/issues/2516

I've verified that the new tests pass in the quintush fork of helm-unittest which we plan on switching to.

### DIFF
--- a/templates/pgbouncer/pgbouncer-networkpolicy.yaml
+++ b/templates/pgbouncer/pgbouncer-networkpolicy.yaml
@@ -1,7 +1,7 @@
 ################################
 ## Pgbouncer NetworkPolicy
 #################################
-{{- if .Values.networkPolicies.enabled }}
+{{- if and .Values.networkPolicies.enabled .Values.pgbouncer.networkPolicies.enabled }}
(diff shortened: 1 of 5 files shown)


templates/pgbouncer/pgbouncer-networkpolicy.yaml
```diff
@@ -1,7 +1,7 @@
 ################################
 ## Pgbouncer NetworkPolicy
 #################################
-{{- if .Values.networkPolicies.enabled }}
+{{- if and .Values.networkPolicies.enabled .Values.pgbouncer.networkPolicies.enabled }}
```
kaxil, line 4: We should probably have `.Values.pgbouncer.enabled` too, right?

```suggestion
{{- if and  .Values.pgbouncer.enabled .Values.networkPolicies.enabled .Values.pgbouncer.networkPolicies.enabled }}
```

kaxil (approved):

PR #205, opened 2021-04-02
Add a doc on how to write helm unitests.

danielhoherd:
Document how to write unittests so we can refer people to that when we ask them to write tests for their submissions, which has been a big turn off to people who are unfamiliar with testing helm charts.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,138 @@
+# Testing
+
+This project uses the [helm-unittest project](https://github.com/quintush/helm-unittest) to test the helm charts included within. Once that plugin is installed, you can run tests with `helm unitest -3 .`
+
+## Writing tests
+
+Tests should be written with one template per file. When writing tests, you need to find content to make assertions against. The best way to do this is something like:
+
+```sh
+helm template . \
+  --set pgbouncer.enabled=True \
+  --set networkPolicies.enabled=True \
+  --set pgbouncer.networkPolicies.enabled=True \
+  --show-only templates/pgbouncer/pgbouncer-networkpolicy.yaml
+```
+
+Which produces the following output:
+
+```yaml
+---
+# Source: airflow/templates/pgbouncer/pgbouncer-networkpolicy.yaml
+################################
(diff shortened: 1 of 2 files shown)


tests/README.md
```diff
@@ -0,0 +1,138 @@
+# Testing
+
+This project uses the [helm-unittest project](https://github.com/quintush/helm-unittest) to test the helm charts included within. Once that plugin is installed, you can run tests with `helm unitest -3 .`
```
kaxil, line 3: We could also move to a more Pythonic way if helm-unittest needs to be forked  https://github.com/apache/airflow/blob/master/TESTING.rst#helm-unit-tests

WDYT? We use it for OSS Airflow and it has worked well for us so far, but I don't have a strong preference here.

Example commit to add a feature and add unit test for it in for OSS Airflow: https://github.com/apache/airflow/commit/441b4ef19f07d8c72cd38a8565804e56e63b543c#diff-f02a725253639ce892b023d3b6e307946244ebcba2b50af02feed402ca80ace9

danielhoherd: Interesting. I had thought about the possibility of testing helm with python but never looked into it. I'll have to think about this some more.

jedcunningham: I think this is a good idea. We should open another issue for it though.

tests/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+From this output, we can make the following assertions:
+
+```yaml
+---
+suite: Test templates/pgbouncer/pgbouncer-networkpolicy.yaml
+templates:
+  - templates/pgbouncer/pgbouncer-networkpolicy.yaml
+tests:
+  - it: "should work with pgbouncer.enabled: True, networkPolicies.enabled: True, pgbouncer.networkPolicies.enabled: True"
```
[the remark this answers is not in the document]
danielhoherd: Yeah, good point. It's something we carried over from the original examples. I'm open to the idea of changing the wording.

kaxil (approved):

Issue #206, opened 2021-04-02
Use helm-doc to generate parameter table


danielhoherd:
We should investigate using [helm-docs](https://github.com/norwoodj/helm-docs) to generate our parameters table, which is out of date and does not include all of our parameters.

PR #207, opened 2021-04-02
Auto-generate helm values in README.md

danielhoherd:
This is a first crack at auto-generating the helm values in the README.md, as described briefly in https://github.com/astronomer/airflow-chart/issues/206. I'm not sure if this is a great idea, so I haven't finished this. I wanted to open this up to discussion before making any more progress.

One thing I'm not happy about is how the table contents are not rendered in an ascii readable way. I experimented with auto-formatting using prettier, but having both helm-docs and prettier modifying the file led to an always-fail scenario.

Another problem is that our values go so deep, the resulting table has horizontal scrolling that hides all but the left-most column. See <https://github.com/astronomer/airflow-chart/blob/7384359163/README.md>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,12 @@
 ---
-exclude: '(venv|.vscode)'  # regex
+exclude: '^(venv|\.vscode)' # regex
 repos:
+  - repo: https://github.com/norwoodj/helm-docs
+    rev: v1.2.0
+    hooks:
+      - id: helm-docs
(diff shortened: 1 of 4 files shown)


kaxil:
I like the approach, I pushed https://github.com/astronomer/airflow-chart/pull/207/commits/5210691162b5946d69c66362cd125868194d7654 to reduce the width:

WDYT now @danielhoherd : https://github.com/astronomer/airflow-chart/blob/5210691162b5946d69c66362cd125868194d7654/README.md#values

PR #215, opened 2021-04-21
Fix passing upgrading strategies to Celery Worker

kaxil:
We create a Statefulset if persistence is enabled and Deployment
otherwise. Upgrading strategies are different for both those types
of resources.

Statefulst needs `updateStrategy`, deployment needs `strategy`.

Without this fix, using Helm Chart with CeleryExecutor failed with

```
helm install airflow -n airflow --set executor=CeleryExecutor --set images.airflow.repository=k8s-dags --set images.airflow.tag=1.0.2 .
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec): unknown field "strategy" in io.k8s.api.apps.v1.StatefulSetSpec
```

## 📋 Checklist

- [x] Helm chart unit test(s)


values.yaml
```diff
@@ -155,11 +155,13 @@ sccEnabled: false
...
-      rollingUpdate:
-        maxSurge: "100%"
-        maxUnavailable: "50%"
+  # Update Strategy when worker is deployed as a StatefulSet
+  updateStrategy: ~
+  # Update Strategy when worker is deployed as a Deployment
+  strategy:
+    rollingUpdate:
+      maxSurge: "100%"
+      maxUnavailable: "50%"
```
kaxil, line 164: This didn't affect us currently as by default persistence is not enabled, hence we deploy a Deployment instead of Statefulset.

jedcunningham (approved):

PR #219, opened 2021-05-20
Add support for using an external Celery broker

jedcunningham:
This allows Redis to be disabled and a separate Celery broker to be
configured in its place with `data.brokerUrl` or
`data.brokerUrlSecretName`.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -127,6 +127,7 @@ data:
   # If secret names are provided, use those secrets
   metadataSecretName: ~
   resultBackendSecretName: ~
+  brokerUrlSecretName: ~
 
   # Otherwise pass connection values in
   metadataConnection:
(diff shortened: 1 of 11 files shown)


values.yaml
```diff
@@ -417,7 +420,6 @@ redis:
 
   # If set use as redis secret
   passwordSecretName: ~
-  brokerURLSecretName: ~
```
jedcunningham, line 420: I removed this since as best as I can tell, it isn't being used by the platform at all and was only used in a truthy check to prevent the `redis-password` Secret from being created. Using `redis.passwordSecretName` alone is sufficient to do this:

https://github.com/astronomer/airflow-chart/blob/f224a0753fdff15b3063694bf1c69850f67273c9/templates/secrets/redis-secrets.yaml#L5

kaxil (approved):

ianstanton (approved):
LGTM

PR #223, opened 2021-05-30
Improve pre-commit. Add ci step.

danielhoherd:
- Improve pre-commit config.
- Add pre-commit to CI.

This deliberately has not been run against the repo yet. There may be some PRs that we want to merge before rebasing against master, then running `pre-commit run --all-files` and fixing all the complaints, which will cause fewer merge conflicts.


.circleci/config.yml
```diff
@@ -44,6 +46,30 @@ workflows:
...
+  run_pre_commit:
+    docker:
+      - image: quay.io/astronomer/ci-pre-commit:2021-05
+    steps:
+      - checkout
+      - run:
+          name: "Create pre-commit-cache-key.txt"
+          command: |
+            cp .pre-commit-config.yaml /tmp/pre-commit-cache-key.txt
+            python --version --version | sed 's/^/# /' >> /tmp/pre-commit-cache-key.txt
```
kaxil, line 58:
```suggestion
            python --version | sed 's/^/# /' >> /tmp/pre-commit-cache-key.txt
```

kaxil (approved):

.circleci/config.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           name: "Create pre-commit-cache-key.txt"
           command: |
             cp .pre-commit-config.yaml /tmp/pre-commit-cache-key.txt
-            python --version --version | sed 's/^/# /' >> /tmp/pre-commit-cache-key.txt
+            python --version | sed 's/^/# /' >> /tmp/pre-commit-cache-key.txt
```
danielhoherd, line 58: @kaxil this needed two `--version` strings:
```
$ python --version
Python 3.8.5
$ python --version --version
Python 3.8.5 (default, Jan 27 2021, 15:41:15)
[GCC 9.3.0]
```

kaxil: oh wow ! Didn't know about it

kaxil: Go for it -- I didn't know. TIL:

```
-V     : print the Python version number and exit (also --version)
         when given twice, print more information about the build
```

danielhoherd: yeah, the syntax is a bit ambiguous. I feel like the extra info should be behind a more descriptive flag like `--build-info`

danielhoherd: https://github.com/astronomer/airflow-chart/pull/224

PR #228, opened 2021-06-03
Only run no-commit-to-branch hook locally

jedcunningham:
We don't need to run no-commit-to-branch on CI as that ends up failing
when we run precommit on master.

Edit: A little more color: That check is really meant to protect against accidents locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,8 @@ jobs:
       - run:
           name: Run pre-commit
           command: pre-commit run --all-files || { git --no-pager diff && false ; }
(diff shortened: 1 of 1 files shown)


kaxil (approved):

danielhoherd (approved):
LGTM. TIL.

danielhoherd:
FWIW [I've added this to the ci image itself](https://github.com/astronomer/ci-images/commit/25a175efdac595473ddadef6c988b565cce94f83) so we won't need to apply this change elsewhere.

PR #231, opened 2021-06-16
Fix labels set by pod_mutation_hook

jedcunningham:
Set the proper values for `kubernetes_executor` and
`kubernetes_pod_operator` labels.

This is critical, especially when using 2.0+, KubernetesExecutor, KPO and
`is_delete_operator_pod=False`, as the `kubernetes_executor` label is used
in the completed pod adoption logic in KubernetesExecutor. With the wrong
labels, the executor tries to adopt the task pod.

Before:
```
task:
  kubernetes_executor=True                                                                                                                                                                                
  kubernetes_pod_operator=False

```

After:
```
task:
  kubernetes_executor=False                                                                    
  kubernetes_pod_operator=True 
```

And in both cases workers get:
```
worker:
  kubernetes_executor=True
  kubernetes_pod_operator=False
```

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -159,7 +159,7 @@ data:
             "kubernetes_pod_operator": "False"
         }
 
-        if 'airflow-worker' in pod.labels.keys() or \
+        if 'airflow-worker' in pod.labels.keys() and \
                 conf.get('core', 'EXECUTOR') == "KubernetesExecutor":
             extra_labels["kubernetes_executor"] = "True"
         else:
@@ -181,7 +181,7 @@ data:
         }
 
 
-        if 'airflow-worker' in pod.metadata.labels.keys() or \
+        if 'airflow-worker' in pod.metadata.labels.keys() and \
                 conf.get('core', 'EXECUTOR') == "KubernetesExecutor":
             extra_labels["kubernetes_executor"] = "True"
         else:


templates/configmap.yaml
```diff
@@ -159,7 +159,7 @@ data:
             "kubernetes_pod_operator": "False"
         }
 
-        if 'airflow-worker' in pod.labels.keys() or \
+        if 'airflow-worker' in pod.labels.keys() and \
```
kaxil, line 162: The problem with this is the old Airflow versions didn't have 'airflow-worker' label IIRC, might have to check.

jedcunningham: I went back to 1.10.10, I can go further though.

jedcunningham: It's in 1.10.0, so I think we should be good.

kaxil (approved):

kaxil:
Looks like this was fixed in https://github.com/astronomer/astronomer/pull/1059 at astronomer/astronomer by @rbankston

kaxil:
Can you PR here it in https://github.com/astronomer/google-environments too: https://github.com/astronomer/google-environments/blob/fb8522ba4e0cca3749ddfd1c37eb8ae4d2e29f79/config.yaml#L248-L249

rbankston:
@kaxil the google-environments should already have that change

jedcunningham:
It doesn't have the `or` -> `and` change, which causes issues.

PR #234, opened 2021-06-28
Drop support for Helm 2

jedcunningham:
Helm 2 is EOL, so bump our chart to the v2 apiVersion

https://helm.sh/blog/helm-v2-deprecation-timeline/


kaxil:
https://app.circleci.com/pipelines/github/astronomer/airflow-chart/965/workflows/e986caa9-4413-45ef-8fbf-fac2aff1bb14/jobs/10963 is failing

```
Installed plugin: unittest

### Error:  apiVersion 'v2' is not valid. The value must be "v1"


Charts:      1 failed, 1 errored, 0 passed, 1 total
Test Suites: 0 passed, 0 total
Tests:       0 passed, 0 total
Snapshot:    0 passed, 0 total
Time:        3.828544ms

Error: plugin "unittest" exited with error
```

jedcunningham:
Yeah, I'm tempted to just table this until the OSS subchart bit lands. Then we can get off the deprecated helm-unittest dependency as we will have way less to convert to pytest.

kaxil:
Sounds good

danielhoherd:
@jedcunningham this error is because you have to pass `-3` to `helm unittest`. EG: https://github.com/astronomer/astronomer/blob/master/Makefile#L37

jedcunningham:
@danielhoherd, ah, good catch. I didn't even think to check that 🤷‍♂️

andriisoldatenko (approved):
yeah LGTM!

andriisoldatenko:
@jedcunningham only i would love to have an issue, so we can track changes.

PR #235, opened 2021-07-15
From oss with sidecar

aliotta:
## Description

## PR Title

## 🎟 Issue(s)

## 🧪  Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)


requirements.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: airflow
-  repository: https://github.com/astronomer/airflow/releases/download/helm-chart%2Fv1.0.1-dev1/
+  repository: https://github.com/astronomer/airflow/releases/download/oss-helm-chart%2F1.1.0-rc1/
```
aliotta, line 3: not sure what this change is from. Should I try w/o it?

kaxil: Yeah, let's use that change

jedcunningham: Yeah, that's the latest snapshot of the OSS chart. We definitely need it.

When the `1.1.0` chart lands, this will switch to `https://airflow.apache.org` (most likely).

values.yaml
```diff
@@ -257,3 +257,10 @@ ingress:
 #  networkPolicies:
 #    enabled: true
 #    # TODO: Originated from https://github.com/astronomer/issues/issues/2516
+
+platform:
+  release: ~
+  workspace: ""
```
jedcunningham, line 263: Is `workspace` actually used?

values.yaml
```diff
@@ -257,3 +257,10 @@ ingress:
...
+ports:
+  flowerUI: 5555
+  airflowUI: 8080
```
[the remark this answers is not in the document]
aliotta: will test now

values.yaml
aliotta: Looks like a no I'll test now

values.yaml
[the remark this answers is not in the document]
aliotta: yep still works. update pushed.

values.yaml
aliotta: yep still works. update pushed.

jedcunningham (approved):

rdodev (approved):

PR #239, opened 2021-08-20
Quick fix to allow postgresql dependency to install

upkash:
## Description

Fixes the repository for the postgresql dependency.

## PR Title

Quick fix to allow postgresql chart to install

## 🧪  Testing

There were no new features added, however the change may not work for all platforms. I tested by deploying airflow to minikube and AWS EKS, and both times it did not work until I made the change submitted in this PR

## 📋 Checklist

- [ x] The PR title is informative to the user experience
- [x] Functional test(s) added
- [ ] Helm chart unit test(s)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ dev-account.json
 kubeconfig
 letsencrypt*
 repository
+requirements.lock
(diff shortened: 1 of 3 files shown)


kaxil:
Thanks for your contribution but this is no longer required from 1.0.0+ version of chart after https://github.com/astronomer/airflow-chart/pull/250

PR #256, opened 2021-10-21
add pod template file override

aliotta:


danielhoherd:
I really wish we had a description for this change... I'm looking for context on parts of it.

PR #260, opened 2021-11-03
Add JSON Schema file

thmzlt:
## Description

For now, we are interested in checking that all OSS chart values are correctly
namespaced under the `airflow` top-level property, so we declare all top-level
properties and set `additionalProperties` to false.

## 🎟 Issue(s)

Resolves astronomer/issues#3715

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/tests/extra-objects_test.yaml
+++ b/tests/extra-objects_test.yaml
@@ -9,9 +9,8 @@ tests:
           count: 0
   - it: should add extra objects
     set:
-      something: world
       extraObjects:
-        - name: "hello {{ .Values.something }}"
+        - name: "hello world"
           another:
             something: yes
     asserts:
--- a/values.schema.json
+++ b/values.schema.json
@@ -1,0 +1,32 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "airflow": {
+      "type": "object"
+    },
+    "authSidecar": {
+      "type": "object"
+    },
+    "extraObjects": {
+      "type": "array"
+    },
+    "helmValuesRevisionId": {
+      "type": "integer"
+    },
+    "ingress": {
+      "type": "object"
+    },
+    "platform": {
+      "type": "object"
+    },
+    "sccEnabled": {
+      "type": "boolean"
+    },
+    "workers": {
+      "type": "object"
+    }
+  },
+  "title": "Values",
+  "type": "object"
+}
--- a/tests/workers_worker-horizontalpodautoscaler_test.yaml
+++ b/tests/workers_worker-horizontalpodautoscaler_test.yaml
@@ -5,7 +5,8 @@ templates:
 tests:
   - it: should work
     set:
-      executor: CeleryExecutor
+      airflow:
+        executor: CeleryExecutor
       workers.autoscaling.enabled: true
     asserts:
       - isKind:


values.schema.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "platform": {
+      "type": "object"
+    },
+    "sccEnabled": {
+      "type": "boolean"
+    },
+    "workers": {
+      "type": "object"
+    }
```
jedcunningham, line 28: It might be worth expanding this to cover everything we currently support, as `workers.extraVolumeMounts` for example would pass validation but no longer work.

danielhoherd: I would say that if `workers.extraVolumeMounts` used to work and after the OSS helm chart switch it does not work, it's worth doing. @jedcunningham do you know if that is the case?

thmzlt: I'll update the schema file to include declaration for all top-level properties (and their child properties) except for `airflow` which is delegated to the OSS chart. Would that work? Any other concerns?

jedcunningham: In this first pass, only `workers` needs to be comprehensive. The others didn't get moved (and in the case of workers, partially moved).

thmzlt: Thanks, @jedcunningham. I added types/properties for everything under `workers`.

aliotta:
While it makes sense to have a JSON schema defined in this repo this does not solve the use case we were looking to resolve. Specifically we wanted a solution that would inform our enterprise users that they have an issue in their environment variables at the time of the upgrade of their platform. The way to solve this is to have a very similar schema file to this defined in the astronomer/astronomer repo.

thmzlt:
@aliotta: https://github.com/astronomer/astronomer/pull/1270

values.schema.json
```diff
@@ -0,0 +1,77 @@
...
+      "type": "boolean"
+    },
+    "workers": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "affinity": {
+          "type": "object"
+        },
+        "autoscaling": {
```
jedcunningham, line 33: `autoscaling` is the only thing that should be here. The others, if they exist, _should_ be an error as they should be under `airflow.workers` instead.

https://github.com/astronomer/airflow-chart/blob/d0c176ddaf42e1bd53e48c1c063057394aee464d/values.yaml#L340-L347

jedcunningham (changes requested):

tests/extra-objects_test.yaml
```diff
@@ -9,9 +9,8 @@ tests:
           count: 0
   - it: should add extra objects
     set:
-      something: world
       extraObjects:
-        - name: "hello {{ .Values.something }}"
+        - name: "hello world"
```
jedcunningham, line 13: Why are we removing the templating part of this test?

jedcunningham (approved):

tests/extra-objects_test.yaml
thmzlt: Because `.Values.something` does not validate against the schema.

On a second thought, I should either remove it altogether or replace with something more useful.

jedcunningham: Ah gotcha, yeah I think it'd make sense to keep testing that templating works if you can come up with something you like. Using the release name would be a good one.

danielhoherd:
@jedcunningham @aliotta what's the status here? Thomaz isn't around to finish this. Do we want to finish this effort or abandon it?

PR #262, opened 2021-11-09
Add pre-commit, tweak CircleCI configs

danielhoherd:
Add pre-commit hooks that are more in line with what we have in astronomer/astronomer. Also add prettier. Run `pre-commit run --all-files` to correct the contents of the repo.

Also bump some dev component versions in CircleCI.

There are no functional changes in this change set. Closest thing is sorting a requirements.txt file. Other than that and pre-commit config, everything else is just markdown changes, and mostly whitespace.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,26 @@
 ---
 exclude: '^(venv|\.vscode)' # regex
 repos:
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.1.10
+    hooks:
+      - id: remove-tabs
+        exclude_types: [makefile, binary]
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: "v2.4.1"
+    hooks:
+      - id: prettier
+        args: ["--print-width=135"]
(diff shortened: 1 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -1,10 +1,26 @@
...
   - repo: https://github.com/psf/black
-    rev: 21.9b0
+    rev: 21.10b0
     hooks:
       - id: black
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.29.0
+    hooks:
+      - id: pyupgrade
+        args: ["--py38-plus"]
```
jedcunningham, line 23: How did you land on python 3.8+?

danielhoherd: This is the version we use for testing the charts. We could drop this to 3.7, but 3.6 would be too far. https://endoflife.date/python

jedcunningham: Gotcha, yeah since it only applies to our testing setup and a sample DAG, this is good.

jedcunningham (approved):

kaxil (approved):

PR #265, opened 2021-11-16
3708 git-sync-relay

danielhoherd:
## Description

Add git-sync-relay features to airflow-chart.

## 🎟 Issue(s)

Resolves astronomer/issues#3708 and https://github.com/astronomer/issues/issues/4059

## 🧪 Testing

git-sync is already working. We need to test that it works exactly the same from the user perspective with git-sync-relay, but all git-sync remotes are the git-sync-relay service, except the relay itself whose remote is the actual git repo being cloned.


shmanu017 (approved):
LGTM

pgvishnuram:
@danielhoherd - I dont see any resource section defined in templates for git-sync

danielhoherd:
@pgvishnuram good catch! We need to put a test into the airflow-chart to check for that in all everything, and add those missing configs.

PR #266, opened 2021-11-16
Fix git sync rev param hash cloning issue

bote795:
## Description

> update git sync version to 3.3.4 from 3.3.0 to fix rev git sync param accepting git hash

looked at the following for reference: https://github.com/apache/airflow/blob/1ee65bb8ae9f98233208ebb7918cf9aa1e01823e/chart/values.yaml#L85 

## 🎟 Issue(s)

Resolves astronomer/issues#3779

## 🧪 Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -39,8 +39,10 @@ airflow:
     pgbouncerExporter:
       repository: quay.io/astronomer/ap-pgbouncer-exporter
       tag: 0.11.0-2
-
-  # Airflow scheduler settings
+    gitSync:
+      repository: k8s.gcr.io/git-sync/git-sync
+      tag: v3.3.4
+    # Airflow scheduler settings
   scheduler:
     livenessProbe:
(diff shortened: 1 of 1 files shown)


aliotta (approved):

values.yaml
```diff
@@ -39,8 +39,10 @@ airflow:
     pgbouncerExporter:
       repository: quay.io/astronomer/ap-pgbouncer-exporter
       tag: 0.11.0-2
-
-  # Airflow scheduler settings
+    gitSync:
+      repository: k8s.gcr.io/git-sync/git-sync
+      tag: v3.3.4
+    # Airflow scheduler settings
```
jedcunningham, line 45:
```suggestion
  # Airflow scheduler settings
```

jedcunningham (approved):

PR #269, opened 2021-11-18
Fix KubernetesExecutor on Airflow < 1.10.12

jedcunningham:
Airflow pre 1.10.12 doesn't use the pod_template_file and we were missing some of the old-style configuration needed to allow KubernetesExecutor to function.

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/3799

## 🧪 Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -327,8 +327,19 @@ airflow:
       elasticsearch_write_stdout: True
       elasticsearch_json_format: True
       elasticsearch_log_id_template: "{dag_id}_{task_id}_{execution_date}_{try_number}"
+    # The following kubernetes config is required to support Airflow 1.10.10
     kubernetes:
       dags_in_image: True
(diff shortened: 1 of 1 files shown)


kaxil (approved):

values.yaml
```diff
@@ -327,8 +327,19 @@ airflow:
       elasticsearch_write_stdout: True
       elasticsearch_json_format: True
       elasticsearch_log_id_template: "{dag_id}_{task_id}_{execution_date}_{try_number}"
+    # The following kubernetes config is required to support Airflow 1.10.10
```
kaxil, line 330: For future readers -- Airflow 2.x will ignore these configs so there are here just for backwards compatibility

kaxil:
I can confirm this works just with the chart -- just tested it

image

image

PR #272, opened 2021-11-23
fix release to prod command for multi version charts

aliotta:
Issue: https://github.com/astronomer/issues/issues/3816

Description: when the chart file has a field called version twice the awk command concats them together. for instance the below chart outputs 1.0.91.3.0 when it should output 1.0.9.

`# apiVersion v2 is Helm 3
apiVersion: v2
name: airflow
version: 1.1.0-rc1
description: Helm chart to deploy the Astronomer Platform Airflow module
icon: https://airflow.apache.org/docs/apache-airflow/stable/_images/pin_large.png
keywords:
  \- astronomer
  \- airflow
dependencies:
  \- name: airflow
    version: 1.3.0
    repository: https://airflow.apache.org`

### DIFF
--- a/bin/release-internal-to-prod
+++ b/bin/release-internal-to-prod
@@ -15,7 +15,7 @@ fi
 dev_repo=internal-helm.astronomer.io
 prod_repo=helm.astronomer.io
 git_root="$(git rev-parse --show-toplevel)"
-version=$(awk '$1 ~ /^version/ {printf $2}' "${git_root}/Chart.yaml")
+version=$(awk '$1 ~ /^version/ {printf $2;exit;}' "${git_root}/Chart.yaml")
 filename="airflow-${version}.tgz"
 max_age=300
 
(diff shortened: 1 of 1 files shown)


bin/release-internal-to-prod
```diff
@@ -15,7 +15,7 @@ fi
 dev_repo=internal-helm.astronomer.io
 prod_repo=helm.astronomer.io
 git_root="$(git rev-parse --show-toplevel)"
-version=$(awk '$1 ~ /^version/ {printf $2}' "${git_root}/Chart.yaml")
+version=$(awk '$1 ~ /^version/ {printf $2;exit;}' "${git_root}/Chart.yaml")
```
aliotta, line 18: This still assumes that the first version param it finds is the one we need.

aliotta: Is it possible to use jq here instead so we can verify its a top level param

danielhoherd (approved):

PR #280, opened 2021-12-29
pgbouncer sslmode support

pgvishnuram:
## Description

Adds support for sslmode "require" when posgtres server is enabled with force tls

## PR Title

pbouncer sslmode support

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/3482

## 🧪 Testing

## 📸 Screenshots

## 📋 Checklist

- [ ] Helm chart unit test(s)

### DIFF
--- a/templates/generate-ssl.yaml
+++ b/templates/generate-ssl.yaml
@@ -1,0 +1,83 @@
+###########################################
+## Pgbouncer SSL Pre install/update Hook ##
+###########################################
+{{- if (and .Values.airflow.pgbouncer.enabled (eq .Values.airflow.pgbouncer.sslmode "require")) }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    "helm.sh/hook": "pre-install,pre-upgrade"
+    "helm.sh/hook-weight": "-2"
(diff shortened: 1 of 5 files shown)


templates/generate-ssl.yaml
```diff
@@ -0,0 +1,83 @@
...
+        tier: airflow
+        component: pgbouncer
+        release: {{ .Release.Name }}
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Release.Name }}-pgbouncer-ssl
+      containers:
+        - name: pgbouncer-certgenerator
+          image: "{{ .Values.astronomer.images.certgenerator.repository }}:{{ .Values.astronomer.images.certgenerator.tag}}"
+          imagePullPolicy: Always
```
danielhoherd, line 75: This should not be set to `Always`
```suggestion
          imagePullPolicy: IfNotPresent
```

danielhoherd (changes requested):
@pgvishnuram All helm related changes we make should have tests, if only to validate that the manifests render correctly with every supported version of k8s. In this case we can have additional tests for custom image name since that is a variable in `values.yaml`

danielhoherd (approved):

PR #282, opened 2022-01-04
Add logging sidecar configmap

thmzlt:
## Description

Creates a fluentd configmap for sidecar logging when the feature is enabled.

## 🎟 Issue(s)

Related to astronomer/issues#3556.


templates/logging-sidecar-configmap.yaml
```diff
@@ -0,0 +1,95 @@
+{{- if .Values.loggingSidecar.enabled }}
```
danielhoherd, line 1: Can you put a file header here like are in many of the other templates? Having one makes helm behave a little nicer when the whole template is wrapped in a check for enabled, causing the template to return some data even when the feature is disabled. Otherwise we can run into errors that are inconsistent between the feature being enabled and disabled.

danielhoherd (changes requested):
Mostly LGTM. Just want to see a header in the file. Also pre-commit is failing on a trivial line-ending change.

templates/logging-sidecar-configmap.yaml
thmzlt: Thanks for the explanation. Added the header comment to the configmap template.

danielhoherd (approved):

PR #290, opened 2022-01-28
Update deprecated api groups for Kubernetes 1.22 support

rbankston:
## Description

Update api versions to be compatible with k8s >= 1.22

## Related Issues

[Part of work to support Kubernetes 1.22](https://github.com/astronomer/issues/issues/3889)

[Astronomer Airflow Chart Issue Link to support 1.22](https://github.com/astronomer/issues/issues/4363)

## Testing

- Initial Validation: Added 1.22 k8s kind cluster to run Airflow deployments 
- CI passes on 1.22 and 1.23

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ workflows:
           matrix:
             parameters:
               # https://hub.docker.com/r/kindest/node/tags
-              kube_version: ["1.19.11", "1.20.7", "1.21.2"]
+              kube_version: ["1.19.11", "1.20.7", "1.21.2", "1.22.5"]
               executor: ["CeleryExecutor", "LocalExecutor", "KubernetesExecutor"]
           requires:
             - build-and-release-internal
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ workflows:
           matrix:
             parameters:
               # https://hub.docker.com/r/kindest/node/tags
-              kube_version: ["1.19.11", "1.20.7", "1.21.2", "1.22.5"]
+              kube_version: ["1.19.11", "1.20.7", "1.21.2", "1.22.7", "1.23.4"]
               executor: ["CeleryExecutor", "LocalExecutor", "KubernetesExecutor"]
           requires:
             - build-and-release-internal
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -46,24 +46,30 @@ spec:
       http:
         paths:
           - path: /{{ .Release.Name }}/airflow
+            pathType: Prefix
             backend:
-              serviceName: {{ .Release.Name }}-webserver
-              {{- if .Values.authSidecar.enabled  }}
-              servicePort: auth-proxy
-              {{- else }}
-              servicePort: airflow-ui
-              {{- end }}
+              service:
+                name: {{ .Release.Name }}-webserver
+                port:
+                  {{- if .Values.authSidecar.enabled  }}
+                  name: auth-proxy
+                  {{- else }}
+                  name: airflow-ui
+                  {{- end }}
     - host: {{- include "airflow_subdomain" . | indent 1 }}
       http:
         paths:
           - path: /
+            pathType: Prefix
             backend:
-              serviceName: {{ .Release.Name }}-webserver
-              {{- if .Values.authSidecar.enabled  }}
-              servicePort: auth-proxy
-              {{- else }}
-              servicePort: airflow-ui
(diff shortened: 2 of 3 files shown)


templates/ingress.yaml
```diff
@@ -68,7 +68,7 @@ spec:
 {{- if and .Values.ingress.enabled (eq .Values.airflow.executor "CeleryExecutor") }}
 ---
 kind: Ingress
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
```
danielhoherd, line 71: We should change this to something like https://github.com/astronomer/astronomer/blob/c57c70823b/templates/_apivers.tpl#L17-L23

rbankston: Makes sense with 1.19 being new

danielhoherd: Yeah. We've run into problems where customers updated their cluster but not their chart. We should be putting in these `useLegacyThing` options anywhere that the old APIs are available https://github.com/astronomer/astronomer/blob/master/templates/_apivers.tpl#L53

pgvishnuram:
@danielhoherd  added k8s  1.22.5 for validation in circleci pipeline

templates/ingress.yaml
rbankston: made that change for Ingress items.

danielhoherd (changes requested):
Looks fine to me. Description needs to be updated though.

templates/ingress.yaml
```diff
@@ -68,7 +72,7 @@ spec:
...
```
jedcunningham, line 75: Should this have the same backcompat for <1.19 (this is also an ingress)?

danielhoherd: [<1.19 is already deprecated in all cloud providers.](https://github.com/astronomer/ops-stuff/blob/726f67f1f4/docs/k8s-version-table.md)

vishwas-astro:
@rbankston @kaxil @jedcunningham do you folks think that if we fix this in Astronomer's airflow chart, customers will be able to install AC on 1.22 K8s environment? I am asking because Disney has a 1.22 on Rancher environment and our solutions team is trying to install AC on it for them.

cc: @cfcastro24

templates/ingress.yaml
jedcunningham: Cool, so then we don't need backcompat on the other ingress, right?

templates/ingress.yaml
```diff
@@ -4,7 +4,11 @@
 {{- if .Values.ingress.enabled }}
   {{ $ctx := . }}
 kind: Ingress
+{{- if semverCompare "< 1.19-0" .Capabilities.KubeVersion.Version -}}
 apiVersion: networking.k8s.io/v1beta1
+{{- else }}
+apiVersion: networking.k8s.io/v1
+{{- end }}
```
jedcunningham, line 11:
```suggestion
apiVersion: networking.k8s.io/v1
```

templates/ingress.yaml
danielhoherd: That's right. IMHO we can drop everything <1.19 including CI tests for those versions.

danielhoherd: Looks like we already dropped <1.19 from CI 👍

templates/ingress.yaml
pgvishnuram: should we drop - there may be some customer still in 1.19 or 1.18 . GKE has removed all deprecated version , but aws still allows users to create cluster from 1.18

danielhoherd: I had a chat with vishnu today about this. I think the way we should do this is to actively drop it. Customers using 1.18 will not likely be running the latest version of airflow. IE: they are not actively updating software, otherwise they would not be running 1.18.

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ workflows:
           matrix:
             parameters:
               # https://hub.docker.com/r/kindest/node/tags
-              kube_version: ["1.19.11", "1.20.7", "1.21.2"]
+              kube_version: ["1.19.11", "1.20.7", "1.21.2", "1.22.5"]
```
danielhoherd, line 24: We should add 1.23 here too since it's already out and is api compatible with 1.22.
```suggestion
              kube_version: ["1.19.11", "1.20.7", "1.21.2", "1.22.7", "1.23.4"]
```

danielhoherd (approved):
This all LGTM. Passing on k8s 1.22 and 1.23 🎉

PR #291, opened 2022-02-09
Add properly quote namespace names

andriisoldatenko:
resolves https://github.com/astronomer/issues/issues/3885

```
astronomer-houston-worker-776c89bfc5-6ztvm houston 2022-02-09T18:37:08 INFO Response from #createDeployment: {"result":{"message":"failed pre-install: warning: Hook pre-install airflow/templates/generate-ssl.yaml failed: RoleBinding in version \"v1beta1\" cannot be handled as a RoleBinding: v1beta1.RoleBinding.Subjects: []v1beta1.Subject: v1beta1.Subject.Namespace: ReadString: expects \" or n, but found 1, error found in #10 byte of ...|mespace\":12345}]}\n|..., bigger context ...|name\":\"test-number-ns-pgbouncer-ssl\",\"namespace\":12345}]}\n|..."},"deployment":{}}
```

### DIFF
--- a/tests/chart_tests/helm_template_generator.py
+++ b/tests/chart_tests/helm_template_generator.py
@@ -59,7 +59,7 @@ def create_validator(api_version, kind, kube_version="1.21.0"):
 
 def validate_k8s_object(instance, kube_version="1.21.0"):
     """Validate the k8s object."""
-    # Skip PostgreSQL chart
+    # Skip PostgreSQL chart because it doesn't pass kubernetes json schema validation
     labels = jmespath.search("metadata.labels", instance) or {}
     if "helm.sh/chart" in labels:
         chart = labels["helm.sh/chart"]
--- a/tests/chart_tests/helm_template_generator.py
+++ b/tests/chart_tests/helm_template_generator.py
@@ -60,7 +60,7 @@ def create_validator(api_version, kind, kube_version="1.21.0"):
 def validate_k8s_object(instance, kube_version="1.21.0"):
     """Validate the k8s object."""
     # Skip PostgreSQL chart
-    labels = jmespath.search("metadata.labels", instance)
+    labels = jmespath.search("metadata.labels", instance) or {}
     if "helm.sh/chart" in labels:
         chart = labels["helm.sh/chart"]
     else:
--- a/tests/chart_tests/helm_template_generator.py
+++ b/tests/chart_tests/helm_template_generator.py
@@ -22,6 +22,7 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any, Dict, Tuple
 
+import jmespath
 import jsonschema
 import requests
 import yaml
@@ -58,6 +59,14 @@ def create_validator(api_version, kind, kube_version="1.21.0"):
 
 def validate_k8s_object(instance, kube_version="1.21.0"):
     """Validate the k8s object."""
+    # Skip PostgreSQL chart
+    labels = jmespath.search("metadata.labels", instance)
+    if "helm.sh/chart" in labels:
+        chart = labels["helm.sh/chart"]
+    else:
+        chart = labels.get("chart")
+    if chart and "postgresql" in chart:
+        return
     validate = create_validator(
         instance.get("apiVersion"), instance.get("kind"), kube_version=kube_version
     )
@@ -93,7 +102,6 @@ def render_chart(
             tmp_file.name,
             "--namespace",
             namespace,
-
         ]
         if show_only:
             if isinstance(show_only, str):
--- a/tests/chart_tests/helm_template_generator.py
+++ b/tests/chart_tests/helm_template_generator.py
@@ -59,7 +59,7 @@ def create_validator(api_version, kind, kube_version="1.21.0"):
 
 def validate_k8s_object(instance, kube_version="1.21.0"):
     """Validate the k8s object."""
-    # Skip PostgreSQL chart because it doesn't pass kubernetes json schema validation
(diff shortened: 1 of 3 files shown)


jedcunningham (approved):

tests/chart_tests/helm_template_generator.py
```diff
@@ -22,6 +22,7 @@
 from tempfile import NamedTemporaryFile
 from typing import Any, Dict, Tuple
 
+import jmespath
```
danielhoherd, line 25: FWIW I used to be so into using jmespath, but ran into some serious behavior weirdness that made me give up on it entirely. I describe the weirdness here: https://danielhoherd.com/tech-notes/jmespath/#quotes-matter

danielhoherd (approved):

tests/chart_tests/helm_template_generator.py
```diff
@@ -58,6 +59,14 @@ def create_validator(api_version, kind, kube_version="1.21.0"):
 
 def validate_k8s_object(instance, kube_version="1.21.0"):
     """Validate the k8s object."""
+    # Skip PostgreSQL chart
```
danielhoherd, line 62: Why do we skip the postgres chart?

andriisoldatenko: @danielhoherd 
```
namespace = '123'

    @pytest.mark.parametrize("namespace", ["abc", "123", "123abc", "123-abc"])
    def test_namespace_names(namespace):
        """Test various namespace names to make sure they render correctly in templates"""
>       render_chart(namespace=namespace)

tests/chart_tests/test_default_chart.py:14: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/chart_tests/helm_template_generator.py:129: in render_chart
    validate_k8s_object(k8s_object, kube_version=kube_version)
tests/chart_tests/helm_template_generator.py:64: in validate_k8s_object
    validate.validate(instance)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Draft7Validator(schema={'$schema': 'http://json/-...a.org/schema#', 'description': 'Secret holds...etSize bytes.', 'pro...t...d#types-kinds', 'enum': ['Secret'], 'type': ['string', 'null']}, ...}, 'type': 'object', ...}, format_checker=None)
args = ({'apiVersion': 'v1', 'data': {'postgresql-password': 'cG9zdGdyZXM='}, 'kind': 'Secret', 'metadata': {'labels': {'app....ame': 'postgresql', 'helm.sh/chart': 'postgresql-10.5.3'}, 'name': 'RELEASE-NAME-postgresql', 'namespace': 123}, ...},)
kwargs = {}, error = <ValidationError: "123 is not of type 'string', 'null'">

    def validate(self, *args, **kwargs):
        for error in self.iter_errors(*args, **kwargs):
>           raise error
E           jsonschema.exceptions.ValidationError: 123 is not of type 'string', 'null'
E           
E           Failed validating 'type' in schema['properties']['metadata']['properties']['namespace']:
E               {'description': 'Namespace defines the space within which each name '
E                               'must be unique. An empty namespace is equivalent to '
E                               'the "default" namespace, but "default" is the '
E                               'canonical representation. Not all objects are '
E                               'required to be scoped to a namespace - the value of '
E                               'this field for those objects will be empty.\n'
E                               '\n'
E                               'Must be a DNS_LABEL. Cannot be updated. More info: '
E                               'http://kubernetes.io/docs/user-guide/namespaces',
E                'type': ['string', 'null']}
E           
E           On instance['metadata']['namespace']:
E               123

venv/lib/python3.9/site-packages/jsonschema/validators.py:247: ValidationErro
```
because of ^

inspired from OSS chart by @jedcunningham 

https://github.com/apache/airflow/blob/07fe9e8c6acf8826ac1dcaf2ad8c94297a0f0c24/chart/tests/helm_template_generator.py#L95-L96

danielhoherd: So this is to work around a casting problem where `str('123')` is turned into `int(123)`. Is there a way to skip only strings that can be cast as ints ?

andriisoldatenko: @jedcunningham any ideas? @danielhoherd i think it's okey to be on the same page with airflow OOS chart, so we can sync up code bases easily.

jedcunningham: I'm not sure I follow what you're suggesting @danielhoherd.

Ultimately opening a PR to bitnami to fix quoting on namespace names is probably the right way to solve this long term.

danielhoherd: I think it's good to keep in step with OSS, but I think we need to document why we are doing things that don't make logical sense without additional context. Having a comment would be good. Also it would be good to open an issue upstream.

andriisoldatenko: @danielhoherd fixed. PTAL

tests/chart_tests/helm_template_generator.py
```diff
@@ -58,6 +59,14 @@ def create_validator(api_version, kind, kube_version="1.21.0"):
 
 def validate_k8s_object(instance, kube_version="1.21.0"):
     """Validate the k8s object."""
+    # Skip PostgreSQL chart because it doesn't pass kubernetes json schema validation
```
danielhoherd, line 62:
```suggestion
    # Skip PostgreSQL chart because it doesn't pass kubernetes json schema validation with all-numeric namespace
```

PR #312, opened 2022-04-06
update sidecar logging template - replace fluentd with vector elasticsearch configmap

rishkarajgi:
## Description

This PR updates the current sidecar logging template by replacing fluentd configmap with a vector configmap for elastic search consumer

## Related Issues
Related astronomer/issues#4419

## Testing
Tested on a kind cluster

### DIFF
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -5,7 +5,7 @@
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: {{ .Release.Name }}-logging-sidecar
+  name: sidecar-logging
   labels:
     tier: airflow
     component: logging-sidecar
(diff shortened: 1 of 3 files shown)


templates/logging-sidecar-configmap.yaml
```diff
@@ -5,94 +5,33 @@
...
+        type: elasticsearch
+        inputs:
+          - generate_syslog
+        mode: bulk
+        compression: none
+        endpoint: "http://{{ .Values.loggingSidecar.host }}:{{ .Values.loggingSidecar.port }}"
+        auth:
+          strategy: "basic"
+          user: {{ .Values.loggingSidecar.user }}
+          password : {{ .Values.loggingSidecar.password }}
```
ajayy004, line 35: We might have to give first preference to global `customLogging` then elasticsearch service running in `astronomer` namespace.

FYI: We're using vector by datadog [here](https://vector.dev/docs/reference/configuration/sinks/elasticsearch/) is how the config will be mapped.

pgvishnuram:
we need to add  loggingSidecar.enabled flag conditions in the template

ajayy004:
> we need to add loggingSidecar.enabled flag conditions in the template

We already had that in the template, let us know if you meant something else.

danielhoherd:
Need to run this through `pre-commit run --all-files` and commit the changes.

values.yaml
```diff
@@ -374,7 +378,8 @@ authSidecar:
   port: 8084
 
 loggingSidecar:
-  enabled: false
+  enabled: true
```
danielhoherd, line 381: The default here should be false, right?

```suggestion
  enabled: false
```

rishkarajgi: Yup, left it for testing purposes. Have reverted it

templates/logging-sidecar-configmap.yaml
rishkarajgi: @pgvishnuram this will be handled in houston itself right? Precedence given to global ES if custom logging enabled over default config?

pgvishnuram: Yes correct. No extra configuration required

aliotta:
We'll want to add some chart tests for these changes

danielhoherd (approved):

aliotta (approved):

PR #315, opened 2022-04-27
Use the long URL since git.io is shutting down.

danielhoherd:
## Description

[git.io is shutting down on Friday](https://github.blog/changelog/2022-04-25-git-io-deprecation/). This change removes the short URL and just uses the long URL. 

## Related Issues

https://github.com/astronomer/issues/issues/4564

## Testing

This is a very small change that is only related to CI. No testing is needed if CI passes.

## Merging

This change should be merged to every branch we support.


danielhoherd:
Git.io is now broken. Builds are no longer showing git.io links, so I'm going to merge this since the old function is no longer working.

PR #321, opened 2022-05-06
Add securityContext for redis sts to run as non-root

danielhoherd:
## Description

Add securityContext for redis sts to run as non-root

## Related Issues

https://github.com/astronomer/issues/issues/3348

## Testing

This includes tests, which should be enough, as long as CI passes.

## Merging

This is just for release 1.6.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 /test-results
 backup
 certs
+charts
 dev-account.json
 kubeconfig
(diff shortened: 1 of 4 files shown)


pgvishnuram:
@danielhoherd - where are you planning to add the securituyContext for redis - through astronomer chart or in airflow chart

PR #332, opened 2022-06-27
added condition based on custom config

rishkarajgi:
## Description

Loads external vector config from secrets if customConfig flag is enabled

## Related Issues

https://github.com/astronomer/issues/issues/4733

## Testing

Do not merge this PR until this text is replaced with details about how these changes were tested.

## Merging

0.29.2

PR #334, opened 2022-06-29
updated condition to load vector configmap

rishkarajgi:
## Description

A change to previous implementation (https://github.com/astronomer/airflow-chart/pull/332) to load entire configmap only if loggingSidecar: customConfig is disabled, else load from secret using configsyncer

## Related Issues

https://github.com/astronomer/issues/issues/4733

## Related PRs
Houston - https://github.com/astronomer/houston-api/pull/1154
Astronomer - https://github.com/astronomer/astronomer/pull/1575

## Testing

Tested on kind cluster

## Merging

0.29.2

### DIFF
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -1,7 +1,7 @@
 ######################################
 ## Logging sidecar configmap        ##
 ######################################
-{{- if .Values.loggingSidecar.enabled }}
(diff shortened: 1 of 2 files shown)


danielhoherd:
@rishkarajgi I'm confused about what's going on here. It looks like you're just deleting all the work you did before. Can you explain a little bit about how the configMap will be loaded? or maybe we can zoom and just catch up on how all this is fitting together. I'm free starting at 1800Z tomorrow (Thursday) if you want to do that.

pgvishnuram (approved):

PR #342, opened 2022-08-01
update logging sidecar template

pgvishnuram:
## Description

Initially we processed all message via single transform with remap , that was causing json errors in multiple components even-though it was not valid issue. 

The change added in this PR will filter logs based on components and process the logs , this has fixed parse error  for  webserver, scheduler and trigger components

## Related Issues

continuation from the below issue https://github.com/astronomer/issues/issues/4648

## Testing

Yet to update

## Merging

cherry-pick to 1.7, 1.6( not yet confirmed)

### DIFF
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -65,10 +65,20 @@ data:
           }
           .offset = to_int(now()) * 1000000000 + to_unix_timestamp(now()) * 1000000
 
+      final_task_log:
+        type: add_fields
+        inputs:
+          - transform_task_log
+        fields:
+          component: "${COMPONENT:--}"
+          workspace: "${WORKSPACE:--}"
+          release: "${RELEASE:--}"
+
+
       transform_remove_fields:
         type: remove_fields
         inputs:
-          - transform_task_log
+          - final_task_log
           - filter_common_logs
         fields:
           - host
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -54,7 +54,6 @@ data:
       transform_task_log:
         type: remap
         inputs:
-          - filter_common_logs
           - filter_task_logs
         source: |-
           # Parse Syslog input. The "!" means that the script should abort on error.
@@ -70,6 +69,7 @@ data:
         type: remove_fields
         inputs:
           - transform_task_log
+          - filter_common_logs
         fields:
           - host
           - file
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -33,7 +33,6 @@ data:
           workspace: "${WORKSPACE:--}"
           release: "${RELEASE:--}"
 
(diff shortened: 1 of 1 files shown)


vishwas-astro:
@pgvishnuram - is there a GH issue for this? If not, please create one.

templates/logging-sidecar-configmap.yaml
```diff
@@ -33,10 +33,28 @@ data:
...
+          type: "vrl"
+          source: '!includes(["worker"], .component)'
+
+      filter_task_logs:
+        type: filter
+        inputs:
+          - transform_syslog
+        condition:
+          type: "vrl"
+          source: 'includes(["worker"], .component)'
```
jedcunningham, line 51: Does this approach also work for the workers log? I'm thinking we might need to be a little more specific with our condition?

pgvishnuram: Yes - i have tested this change

pgvishnuram: transform_webserver:
        type: filter
        inputs:
          - log_files
        condition: '.component == "worker"'

pgvishnuram: with vrl
```
transform_worker:
        type: filter
        inputs:
          - transform_syslog
        condition: 
          type: "vrl"
          source: '.component == "worker"'
```

jedcunningham: Had a chat with @pgvishnuram and both worker logs and task logs need to follow the 'task log' path here because they are both in json format.

templates/logging-sidecar-configmap.yaml
```diff
@@ -33,10 +33,28 @@ data:
           workspace: "${WORKSPACE:--}"
           release: "${RELEASE:--}"
 
+
```
danielhoherd, line 36:
```suggestion

```

templates/logging-sidecar-configmap.yaml
```diff
@@ -33,10 +33,28 @@ data:
...
+      filter_task_logs:
+        type: filter
+        inputs:
+          - transform_syslog
+        condition:
+          type: "vrl"
+          source: 'includes(["worker"], .component)'
+
+
```
danielhoherd, line 53:
```suggestion

```

templates/logging-sidecar-configmap.yaml
```diff
@@ -47,10 +65,21 @@ data:
...
+      final_task_log:
+        type: add_fields
+        inputs:
+          - transform_task_log
+        fields:
+          component: "${COMPONENT:--}"
+          workspace: "${WORKSPACE:--}"
+          release: "${RELEASE:--}"
+
+
```
danielhoherd, line 77:
```suggestion

```

danielhoherd (approved):

vishwas-astro:
Can we get an approval from @jedcunningham and @danielhoherd before the merge please?

danielhoherd (approved):

PR #364, opened 2022-10-18
add git-secret only if the sshPrivateKeySecretName provided

ajayy004:
## Description

ssh-secret not required when sshPrivateKeySecretName not provided (houston only sends sshPrivateKeySecretName when sshKey config set )

### Related
- https://github.com/astronomer/issues/issues/4059

### DIFF
--- a/tests/chart_tests/test_git_sync_relay_deployment.py
+++ b/tests/chart_tests/test_git_sync_relay_deployment.py
@@ -123,7 +123,7 @@ class TestGitSyncRelayDeployment:
         ]
 
     def test_gsr_deployment_without_ssh_credentials_and_known_hosts(self, kube_version):
-        """Test that a valid deployment is rendered when enabling git-sync with ssh credentials and known hosts and other custom configs."""
+        """Test that a valid deployment is rendered when enabling git-sync without ssh credentials."""
         values = {
             "gitSyncRelay": {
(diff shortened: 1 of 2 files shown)


tests/chart_tests/test_git_sync_relay_deployment.py
```diff
@@ -121,3 +121,60 @@ def test_gsr_deployment_with_ssh_credentials_and_known_hosts(self, kube_version)
             "--wait=333",
             "--root=/git",
         ]
+
+    def test_gsr_deployment_without_ssh_credentials_and_known_hosts(self, kube_version):
+        """Test that a valid deployment is rendered when enabling git-sync with ssh credentials and known hosts and other custom configs."""
```
danielhoherd, line 126: I don't think this test is doing what we need it to do. EG: it doesn't test that there is no secret.

danielhoherd: Wait, I see that it actually is checking that now.

danielhoherd, line 126:
```suggestion
        """Test that a valid deployment is rendered when enabling git-sync without ssh credentials."""
```

ajayy004: Fixed

PR #365, opened 2022-10-21
chart version 1.7.1

pgvishnuram:
## Description

- airflow-chart version 1.7.1

## Related Issues

NA

## Testing

NA

## Merging

Merge to release-1.7 only **after 0.31.0 and 0.30.3 are released**.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v2 is Helm 3
 apiVersion: v2
 name: airflow
-version: 1.7.0
+version: 1.7.1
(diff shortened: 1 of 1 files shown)


danielhoherd (approved):

danielhoherd (changes requested):
We should switch the dependency back to https://airflow.apache.org

danielhoherd (approved):

pgvishnuram:
@danielhoherd - checked with @cmarteepants it can be done on 0.31 since we dont know what extra features added along with that and needs one more regression as well

@vishwas-astro

PR #366, opened 2022-10-21
5145 git-daemon healthcheck

danielhoherd:
## Description

- Add a livenessProbe health check to the git-daemon container to make sure it is exporting the git repo.
- Use latest git-daemon and git-sync image versions.

## Related Issues

https://github.com/astronomer/issues/issues/5145

## Testing

We should test this in a cluster by deleting the `/git/.git/git-daemon-export-ok` file and making sure the healthcheck puts it back. The healthcheck will fail if it is unable to touch the file, which serves as double duty for auto-healing and also alerting when it cannot achieve a healthy state.

Edit: Verified that this new livenessProbe is working as expected, keeping the git dir exported.

Edit: Verified that the new images with non-verbose work as expected.

## Merging

This should be merged into all branches that support git-sync-relay.

### DIFF
--- a/templates/git-sync-relay/git-sync-relay-deployment.yaml
+++ b/templates/git-sync-relay/git-sync-relay-deployment.yaml
@@ -93,4 +93,11 @@ spec:
           env:
             - name: GIT_ROOT
               value: /git
+          livenessProbe:
+            exec:
+              command:
+              - touch
+              - /git/.git/git-daemon-export-ok
+            initialDelaySeconds: 5
+            periodSeconds: 15
 {{- end -}}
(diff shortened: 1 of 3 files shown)


templates/git-sync-relay/git-sync-relay-deployment.yaml
```diff
@@ -93,4 +93,11 @@ spec:
           env:
             - name: GIT_ROOT
               value: /git
+          livenessProbe:
+            exec:
+              command:
+              - touch
+              - /git/.git/git-daemon-export-ok
+            initialDelaySeconds: 5
+            periodSeconds: 15
```
danielhoherd, line 102: I know this is only a single filesystem operation, but maybe this is too frequent?

danielhoherd: Yeah, this isn't going to matter at all with a 1s local git sync period.

bote795 (approved):

PR #375, opened 2022-11-28
Remove deprecated transforms

danielhoherd:
## Description

Replace transforms that were deprecated in vector 0.24 with new syntaxes.

## Related Issues

https://github.com/astronomer/issues/issues/5253

## Testing

I launched a deployment in qa.link and reproduced the problem, then upgraded astronomer to use the airflowChartVersion from this branch, re-deployed, and things worked again. My testing only made sure the pods came up, not that the logs actually exist, because I am not an astronomer admin in the qa.link astronomer deployment that I was testing in.

## Merging

AFAIK these changes are safe to merge anywhere, because these syntaxes have been deprecated for a while and should work with older versions of vector that we have. They are only really needed in release-1.7 though.

### DIFF
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -25,14 +25,14 @@ data:
         read_from: beginning
     transforms:
       transform_syslog:
-        type: add_fields
+        type: remap
         inputs:
           - airflow_log_files
-        fields:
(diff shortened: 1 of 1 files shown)


pgvishnuram (commented):
verified this changes we can safely merge the config

pgvishnuram (approved):

PR #380, opened 2022-12-19
Improve docs to avoid confusion related to the core git-sync feature

danielhoherd:
## Description

Improve docs to avoid confusion related to the core git-sync feature, and other small improvements. This is in response to the issue in this thread:

- <https://astronomer.slack.com/archives/C02NZELT2/p1671464975671469>
- <https://apache-airflow.slack.com/archives/C03T0AVNA6A/p1671450081264109>

## Related Issues

N/A

## Testing

Only docs changes, no testing is needed. I did set up a kind cluster and test these instructions out while changing them.

Because these are only docs changes, the only CI step that needs to pass is run_pre_commit

## Merging

Merge everywhere git-sync-relay is supported.

### DIFF
--- a/docs/git-sync-relay.md
+++ b/docs/git-sync-relay.md
@@ -13,7 +13,7 @@ When using this chart outside of Astronomer, for instance when testing or develo
 First, create an ssh key that has no passphrase:
 
 ```sh
-ssh-keygen -P '' -t ed25519 -f ssh-sync-key-id -C "test key $USER@$HOSTNAME $(date +%FT%T%z)"
+ssh-keygen -P '' -t ed25519 -f airflow-git-ssh -C "test key $USER@$HOSTNAME $(date +%FT%T%z)"
 ```
 
 ### Create a k8s secret from the private key
@@ -21,12 +21,12 @@ ssh-keygen -P '' -t ed25519 -f ssh-sync-key-id -C "test key $USER@$HOSTNAME $(da
 We create a k8s generic secret where the key is stored under `data.gitSshKey`, which is the required location for the private key:
 
 ```sh
-kubectl create secret generic my-secret --from-file=gitSshKey=ssh-sync-key-id
+kubectl create secret generic git-ssh-private-key --from-file=gitSshKey=airflow-git-ssh
 ```
 
 ### Add the pub key to your repo
 
-Take the `ssh-sync-key-id.pub` contents and add it to your https://github.com/settings/keys or whatever server you're cloning from.
+Take the `airflow-git-ssh.pub` contents and add it to your https://github.com/settings/keys or whatever the equivalent is on the server you're cloning from.
 
 ### Install airflow-chart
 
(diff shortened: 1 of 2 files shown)


jedcunningham (approved):

docs/git-sync-relay.md
```diff
@@ -36,18 +36,18 @@ Create a `values.yaml` file with contents similar to the following:
...
     wait: 60 # seconds between synchronizations with upstream git repo
     subPath: dags # if your dags dir is not the repo root, specify the path relative to the repo root
-    sshPrivateKeySecretName: theGitSshPrivateKeyName # The name of a secret that holds the private key.
+    sshPrivateKeySecretName: git-ssh-private-key # This is the secret we created earlier. This is not requred with https git remotes.
 ```
 
 ### Install airflow
 
-Then install airflow like you normally would. Assuming your cwd is the root dir or the astronomer/airflow-chart repo:
+You must also follow the git-sync configuration setup that you would normally follow to install the apache airflow chart. The one customization you must make is to use the git repo served by the git-sync-relay's git-daemon pod as your git remote. That URL is `git://airflow-git-sync-relay.${NAMESPACE}.svc.cluster.local./git` where `${NAMESPACE}` is the airflow installation namespace. Once you have those values set up in your values.yaml file, you can install the astronomer airflow-chart:
```
jedcunningham, line 49: An example might be helpful. Is there any auth on the relay?

danielhoherd: git-daemon serves using git protocol, which has no auth and is quite performant. security is maintained by keeping the git-daemon restricted to the local namespace.

What kind of example are you thinking of?

jedcunningham: Cool. I meant an example values, something like this maybe:

```
airflow:
  dags:
    gitSync:
      enable: true
      repo: git://airflow-git-sync-relay/git`
```

And to make it simpler, just using the service name should be okay right?

```suggestion
You must also follow the git-sync configuration setup that you would normally follow to install the apache airflow chart. The one customization you must make is to use the git repo served by the git-sync-relay's git-daemon pod as your git remote. That URL is `git://airflow-git-sync-relay/git`. Once you have those values set up in your values.yaml file, you can install the astronomer airflow-chart:
```

danielhoherd: @jedcunningham can you check the current version and see if it's suitable?

docs/git-sync-relay.md
```diff
@@ -36,18 +36,28 @@ Create a `values.yaml` file with contents similar to the following:
...
+
+airflow:
+  dags:
+    gitSync:
+      enabled: True
+      repo: git://airflow-git-sync-relay.${NAMESPACE}.svc.cluster.local./git
+      branch: main
+  dagDeployment:
+    repositoryUrl: https://github.com/astronomer/2-4-example-dags
+    type: git_sync
```
jedcunningham, line 54: This must be in the wrong place, that isn't in the OSS chart.

docs/git-sync-relay.md
jedcunningham: Sorry, missed this! Looks good.

docs/git-sync-relay.md
danielhoherd: That's bizarre. I did a deploy in dev and inspected the values and that's the structure I saw, so I configured it and it did work. I'm going to have to go back and look again and figure out wtf is going on because I don't see dagDeployment in either this chart or the OSS chart.

danielhoherd: Oh, my guess is that houston is storing state in the helm values. I'm doing another test now to see what the bare minimum instructions we need to include in doc to get it working are.

danielhoherd: I don't know what I was doing wrong before that was causing a crashloop, but when I try your example today it's working so yay.

jedcunningham: The best bugfix: time 🙃

PR #326, opened 2022-05-27
Fix Elasticsearch log_id_template

jedcunningham:
## Description

Turns out it was `elasticsearch_log_id_template` until 1.10.4, then was
deprecated until 2.0.0, when even backcompat was removed. 2+ relied on
the OSS chart's `log_id_template`, which matched our overridden value.

## Related Issues

https://github.com/astronomer/issues/issues/4634

## Testing

Do not merge this PR until this text is replaced with details about how these changes were tested.

## Merging

I guess `release-1.6` will eventually be merged into `main`? I'm out of the loop on the branching/merging strategy here.

PR #385, opened 2023-01-04
Fix Elasticsearch log_id_template (#326)

pgvishnuram:
## Description

Turns out it was `elasticsearch_log_id_template` until 1.10.4, then was deprecated until 2.0.0, when even backcompat was removed. 2+ relied on the OSS chart's `log_id_template`, which matched our overridden value.

## Related Issues

https://github.com/astronomer/issues/issues/4634

## Testing

No testing needed. This was already merged to release-1.7

## Merging

This is only meant for master.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -338,6 +338,10 @@ airflow:
       write_stdout: True
       elasticsearch_write_stdout: True
       elasticsearch_json_format: True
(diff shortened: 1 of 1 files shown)


danielhoherd (changes requested):
Code changes LGTM. PR description needs to be completed.

jedcunningham:
(This is the same as #326, just to main instead of a release branch)

danielhoherd (approved):

PR #390, opened 2023-01-24
removing worker hpa components

shmanu017:
## Related Issues

https://github.com/astronomer/issues/issues/5373


templates/workers/worker-horizontalpodautoscaler.yaml
```diff
@@ -1,34 +0,0 @@
...
-    heritage: {{ .Release.Service }}
-{{- with .Values.airflow.labels }}
-{{ toYaml . | indent 4 }}
-{{- end }}
-spec:
-  scaleTargetRef:
-    apiVersion: apps/v1
-    kind: StatefulSet
-    name: {{ .Release.Name }}-worker
-  minReplicas: {{ .Values.workers.autoscaling.minReplicas }}
```
jedcunningham, line 23: You need to remove this stuff from values as well.

jedcunningham (changes requested):
There is also a section about this in the README, and you should document this in UPDATING too. Probably should be a major version bump since it is removing a feature?

templates/workers/worker-horizontalpodautoscaler.yaml
danielhoherd: Keda was never shipped as a feature in Astronomer. This is effectively removing dead code that was always dead. I suppose it is possible that folks are using this chart outside of astronomer though...

shmanu017: README , featuer flags and values.yaml cleanup with KEDA and hpa is done

danielhoherd (approved):

danielhoherd (approved):
LGTM

danielhoherd:
I merged this and cherry-picked this to other branches, but then realized that that was not a great plan so I reset those branches to what they were before. I'm on the fence about bumping to a new major or minor version for this breaking change.

pgvishnuram:
@shmanu017 and @danielhoherd 
```
apiVersion: autoscaling/v2beta1
kind: HorizontalPodAutoscaler
```

this is not done keda auto scaler this is from kubernetes metrics server and has no relation to keda autoscaling

shmanu017:
Since , at the moment there is no public facing document.Next step would be deciding either:

-  Create a User facing document with details to use for kubernetes worker metrics. (Since it was added as a feature but never used for last 3 yrs)
```
2019-12-13T13:10:41-05:00 634769b 3 years, 1 month ago <sjmiller609> - Reduced metrics collection for Prometheus (#277)
2019-11-05T10:22:03-05:00 44c3f61 3 years, 3 months ago <sjmiller609> - CPU Throttling alarms (#226)$
```

-  Remove it from airflow-chart (which is the current state in master)

@pgvishnuram @danielhoherd

danielhoherd:
FTR we decided to bump airflow-chart to 1.8 for the HPA removal, fix HPA's k8s 1.25 compatibility in 1.7, and support 1.25 in that older release branch. So we will have:

- 1.25 support everywhere
- HPA removal in airflow-chart 1.8

PR #408, opened 2023-03-10
Improve vector log parsing

danielhoherd:
## Description

- Only parse json lines as json, pass all other messages straight through.
- Rename transform_syslog to transform_airflow_logs to be more accurate

## Related Issues

- <https://github.com/astronomer/issues/issues/5060>

## Testing

I have tested this in aws with kubernetes executor and verified that the vector parse error messages no longer show up. 

## Merging

This should be merged everywhere.

### DIFF
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -24,7 +24,7 @@ data:
           - "${SIDECAR_LOGS}/*.log"
         read_from: beginning
     transforms:
-      transform_syslog:
+      transform_airflow_logs:
         type: remap
         inputs:
           - airflow_log_files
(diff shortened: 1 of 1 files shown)


templates/logging-sidecar-configmap.yaml
```diff
@@ -24,44 +24,44 @@ data:
...
+      transform_airflow_logs:
         type: remap
         inputs:
           - airflow_log_files
         source: |
           .component = "${COMPONENT:--}"
           .workspace = "${WORKSPACE:--}"
           .release = "${RELEASE:--}"
-          .date_nano = parse_timestamp!(.@timestamp, format: "%Y-%m-%dT%H:%M:%S.%f%Z")
+          .date_nano = parse_timestamp!(.@timestamp, format: "%Y-%m-%dT%H:%M:%S.%f%Z") ?? now()
```
danielhoherd, line 35: I'm not sure this is needed. Maybe we can put this one back to the way it was before. What do you think @pgvishnuram ?

rishkarajgi (approved):

PR #409, opened 2023-03-16
5476 statsd configmap

danielhoherd:
## Description

Add the ap-vendor statsd config into this chart as a helm value that is passed to the OSS chart which overrides their statsd config with ours.

## Related Issues

- <https://github.com/astronomer/issues/issues/5476>

## Testing

I've implemented functional and unit tests that should cover all testing needs.

## Merging

This should be merged into all supported branches.


rbankston (approved):
Read through the change and like the simpleness of including it in the values file so if a customer wanted they could override the value through helm.airflow.statsd if desired

pgvishnuram (approved):

PR #440, opened 2023-07-18
Chart version 1.9.0 with apache/airflow chart 1.11.0-astro

danielhoherd:
## Description

- Use apache/airflow chart version 1.11.0 with cron job template change
- Bump to version 1.9.0 because of the big dependency jump (our chart version does not correlate to the OSS chart version)

## Related Issues

https://github.com/astronomer/issues/issues/5670

## Testing

We will need to be on the lookout for changes in airlfow deployments. I have looked at the change list between 1.8.0 and 1.11.0-astro and nothing big jumped out at me, but we should be diligent since we have been bitten by OSS airflow chart changes in the past.

## Merging

This should be merged into 0.33.0, probably 0.32.0 too.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v2 is Helm 3
 apiVersion: v2
 name: airflow
-version: 1.8.7
+version: 1.9.0
 description: Helm chart to deploy the Astronomer Platform Airflow module
 icon: https://airflow.apache.org/docs/apache-airflow/stable/_images/pin_large.png
 keywords:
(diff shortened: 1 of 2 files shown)


Chart.yaml
```diff
@@ -1,13 +1,13 @@
...
 icon: https://airflow.apache.org/docs/apache-airflow/stable/_images/pin_large.png
 keywords:
   - astronomer
   - airflow
 dependencies:
   - name: airflow
-    version: 1.8.0
-    repository: https://airflow.apache.org
+    version: 1.11.0-astro
+    repository: https://github.com/astronomer/airflow/releases/download/oss-helm-chart/1.11.0-astro
```
pgvishnuram, line 13: why are we marking it to internal repo

danielhoherd: @pgvishnuram OSS airflow has not released a chart that has the fix we need. i have released a version of the OSS airflow chart that has the fix. when the public chart catches up, we will switch back. this is something we have done in the past. https://github.com/astronomer/airflow-chart/commit/1e5e98243b5b2c45dd350c9919c3998383a86561

pgvishnuram (approved):
Mostly from QA basic sanity tests and validation.

PR #452, opened 2023-08-22
Test cronjob behavior

danielhoherd:
## Description

Add a test that ensures behavior of the apache/airflow chart that we rely on continues to exist. (beyonce rule)

Also default to using 1.27.0 as kubernetes version.

## Related Issues

N/A

## Testing

This adds a test and modifies existing behaviors. No need for additional testing.

## Merging

This should be merged anywhere that we use apache/airflow chart >=1.11.0.


danielhoherd:
Merged and cherry-picked to 1.8, 1.9

PR #467, opened 2023-11-16
add sidecar logging index prefix overrides

pgvishnuram:
## Description

This PR adds supports to override default index prefix for sidecar logging. Allowing customisation for users to choose their own index prefix.

Note: This does not migrate existing index automatically users should take responsible to migrate their old index data to newer shifted ones.

## Related Issues

https://github.com/astronomer/issues/issues/5947

## Testing

QA should able to pass custom indexPrefix for sidecar logging

## Merging

to all valid releases


danielhoherd (approved):
Looks great!

PR #473, opened 2023-12-06
Add hook that runs 'helm dependency update' when Chart.yaml has changed.

danielhoherd:
## Description

Add pre-commit hook that runs `helm dependency update` when Chart.yaml has changed.

## Related Issues

N/A

## Testing

No testing is needed. This is just a pre-commit change, and a lockfile change created by the new pre-commit hook. These changes don't alter anything that runs in customer environments. The only changes would be when we package the chart, and even that would be a no-op since the packaging scripts already run `helm dependency update`

## Merging

Merge everywhere.


pgvishnuram (approved):

danielhoherd:
Merged and cherry-picked to 1.8, 1.9

PR #474, opened 2023-12-06
Move codeowners to an org team

danielhoherd:
## Description

Move the CODEOWNERS folks out of the file and into a GH org team.

## Related Issues

N/A

## Testing

N/A, this is just repo maintenance

## Merging

Only the main branch is used for the CODEOWNERS feature. We might as well merge everywhere though.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @danielhoherd @aliotta @jedcunningham @pgvishnuram
+* @astronomer/maintainers-airflow-chart


danielhoherd:
The current status here says that the new syntax does not work. This could be because the team has not been approved by IT yet, or because the team is not publicly visible. We need to wait for that syntax error to be resolved. It shows up as a squiggly line underneath the new changes in the PR diff view.

danielhoherd:
I logged in and approved the new team, and verified that it was working, then merged the changes.

PR #482, opened 2024-02-01
initial authsidecar config for dag-server

pgvishnuram:
## Description

This PR adds support for authsidecar configuration in airflow chart in order to support external ingress controller

## Related Issues

https://github.com/astronomer/issues/issues/6123

## Testing

QA should able to access when external ingress controller is used

## Merging

release-0.34

### DIFF
--- a/tests/chart_tests/test_auth_sidecar.py
+++ b/tests/chart_tests/test_auth_sidecar.py
@@ -8,17 +8,15 @@ from . import get_containers_by_name
 
 @pytest.mark.parametrize("kube_version", supported_k8s_versions)
 class TestAuthSidecar:
+    show_only = [
+        "templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml",
+        "templates/flower/flower-auth-sidecar-configmap.yaml",
+        "templates/webserver/webserver-auth-sidecar-configmap.yaml",
+    ]
+
     def test_auth_sidecar_config_defaults(self, kube_version):
         """Test logging sidecar config with defaults"""
-        docs = render_chart(
-            kube_version=kube_version,
-            values={},
-            show_only=[
-                "templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml",
-                "templates/flower/flower-auth-sidecar-configmap.yaml",
-                "templates/webserver/webserver-auth-sidecar-configmap.yaml",
-            ],
-        )
+        docs = render_chart(kube_version=kube_version, values={}, show_only=self.show_only)
         assert len(docs) == 0
 
     def test_auth_sidecar_config_enabled_with_celeryexecutor(self, kube_version):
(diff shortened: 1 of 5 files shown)


danielhoherd (changes requested):
Code looks good, but it needs tests. It looks like we may not have any tests for auth-sidecar. It would be awesome if we had one test file that tests that the feature is applied consistently across all entities that need it.

tests/chart_tests/test_auth_sidecar.py
```diff
@@ -0,0 +1,95 @@
...
+        docs = render_chart(
+            kube_version=kube_version,
+            values={},
+            show_only=[
+                "templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml",
+                "templates/flower/flower-auth-sidecar-configmap.yaml",
+                "templates/webserver/webserver-auth-sidecar-configmap.yaml",
+            ],
+        )
+        assert len(docs) == 0
```
danielhoherd, line 22: We should re-use the show_only list since it is always the same:

```suggestion
@pytest.mark.parametrize("kube_version", supported_k8s_versions)
class TestAuthSidecar:
    show_only = [
        "templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml",
        "templates/flower/flower-auth-sidecar-configmap.yaml",
        "templates/webserver/webserver-auth-sidecar-configmap.yaml",
    ]

    def test_auth_sidecar_config_defaults(self, kube_version):
        """Test logging sidecar config with defaults"""
        docs = render_chart(kube_version=kube_version, values={}, show_only=self.show_only)
        assert len(docs) == 0
```

tests/chart_tests/test_auth_sidecar.py
```diff
@@ -0,0 +1,95 @@
...
+            "subPath": "nginx.conf",
+        }
+
+        authSidecarServicePorts = {
+            "name": "auth-proxy",
+            "port": 8084,
+            "protocol": "TCP",
+            "targetPort": 8084,
+        }
+        docs = render_chart(
```
danielhoherd, line 72: Whitespace nit
```suggestion

        docs = render_chart(
```

tests/chart_tests/test_auth_sidecar.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+        assert len(docs) == 2
+
+    def test_auth_sidecar_config_with_dag_server_enabled(self, kube_version):
+        """Test logging sidecar config with defaults"""
+        resources = {
+            "requests": {"cpu": 99.9, "memory": "777Mi"},
+            "limits": {"cpu": 66.6, "memory": "888Mi"},
+        }
+        volumeMounts = {
```
danielhoherd, line 60: Whitespace nit
```suggestion

        volumeMounts = {
```

danielhoherd (approved):
Looks great. I left some optional suggestions.

PR #488, opened 2024-02-07
Move 'default_nginx_settings' out of /auth

danielhoherd:
## Description

Move 'default_nginx_settings' out of /auth

## Related Issues

https://github.com/astronomer/issues/issues/6140

## Testing

Unfortunately we only have very basic tests for nginx. We'll need to check the symptom in 6140 and see if this fixes it.

## Merging

Merge everywhere since the cause of 6140 was a bug that was fixed in this PR.


pgvishnuram:
I might need to test this again to validate

PR #496, opened 2024-05-13
support pod securitycontext and container security context

pgvishnuram:
## Description

This PR segregates podSecurityContext and containerSecurityContext in order to add changes independently

## Related Issues

Related https://github.com/astronomer/issues/issues/6729

## Testing

QA should able to pass both container securityContext and podSecurity context independently like:
```
shubhamsingh@Shubhams-MacBook-Pro airflow-chart % helm template --kube-version 1.29.0 release-name /Users/shubhamsingh/Desktop/airflow-chart --values values.yaml --show-only templates/dag-deploy/dag-server-statefulset.yaml --set dagDeploy.enabled=true --set dagDeploy.securityContexts.pod.runAsUser=12345 --set dagDeploy.securityContexts.pod.privileged=true --set dagDeploy.securityContexts.container.allowPrivilegeEscalation=false --debug
install.go:222: [debug] Original chart version: ""
install.go:239: [debug] CHART PATH: /Users/shubhamsingh/Desktop/airflow-chart

---
# Source: airflow/templates/dag-deploy/dag-server-statefulset.yaml
############################
## dag-server StatefulSet ##
############################
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: release-name-dag-server
  labels:
    component: dag-server
    tier: airflow
    release: release-name
    chart: "release-name-1.13.3"
    heritage: Helm
spec:
  replicas: 1
  selector:
    matchLabels:
      component: dag-server
      tier: airflow
      release: release-name
  serviceName: release-name-dag-server
  template:
    metadata:
      labels:
        component: dag-server
        tier: airflow
        release: release-name
        chart: "release-name-1.13.3"
        heritage: Helm
    spec:
      serviceAccountName: release-name-dag-server
      securityContext: 
        privileged: true
        runAsUser: 12345
      containers:
        - name: dag-server
          image: "quay.io/astronomer/ap-dag-deploy:0.5.1"
          imagePullPolicy: 
          command: ["sanic", "dag_deploy.server.app", "-H", "0.0.0.0"]
          
          ports:
          - name: server
            containerPort: 8000
          securityContext: 
            allowPrivilegeEscalation: false
          resources:
            {}
          env:
          - name: HOUSTON_SERVICE_ENDPOINT
            value: "http://-houston..svc.cluster.local.:8871/v1/"
          livenessProbe:
            failureThreshold: 5
            httpGet:
              path: /healthz
              port: 8000
              scheme: HTTP
            initialDelaySeconds: 10
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 5
          readinessProbe:
            httpGet:
              path: /healthz
              port: 8000
            initialDelaySeconds: 10
            periodSeconds: 10
          volumeMounts:
          - name: data
            mountPath: /data
            readOnly: false
      volumes:
  volumeClaimTemplates:
  - metadata:
      name: data
    spec:
      accessModes: ["ReadWriteOnce"]
      resources:
        requests:
          storage: 10Gi
shubhamsingh@Shubhams-MacBook-Pro airflow-chart % 

```

## Merging

0.36


templates/dag-deploy/dag-server-statefulset.yaml
```diff
@@ -50,6 +50,7 @@ spec:
           ports:
           - name: server
             containerPort: {{ .Values.dagDeploy.ports.dagServerHttp }}
+          securityContext: {{ toYaml .Values.dagDeploy.securityContext | nindent 12 }}
```
danielhoherd, line 53: Should this be containerSecurityContext ?

danielhoherd: It seems like `securityContext` should apply to both pods and containers, `podSecurityContext` should apply to pods, and `containerSecurityContext` should apply to containers.

pgvishnuram: Yes we can set containerSecurityContext for better naming and we will also release a houston api changes as well for adopting new naming. should we target this release or next patch

danielhoherd: It seems like we should do it on this patch because otherwise customers could put a `securityContext` config in place for containers, and later that becomes a global setting for pods and containers, which is probably not what they intended.

pgvishnuram: @danielhoherd can we follow oss approach here 

```
securityContexts: 
    pods: {}
    container: {}
```

pgvishnuram: so all airflow chart components will be having same approach every where

values.yaml
```diff
@@ -570,9 +570,10 @@ dagDeploy:
   # when it wants to scale a node down.
   safeToEvict: true
 
-  securityContext: {}
+  podSecurityContext: {}
   #  runAsUser: 999
   #  runAsGroup: 0
+  containerSecurityContext: {}
```
pgvishnuram, line 576:
```
securityContexts: 
    pod: {}
    container: {}
```

templates/dag-deploy/dag-server-statefulset.yaml
danielhoherd: i like that!

pgvishnuram: they have top level securityContext and then at service levels - but our use case it to allow passing securityContext when a user what to set more extensive properties at container level

danielhoherd: I think we should probably widen this securityContext effort out into a ticket that standardizes on a way that we will do securityContexts, describe the way we think is best to do them, and then make an effort to do that everywhere. We don't want to come up with a cool way to do this in this one feature if it is inconsistent with how we do it elsewhere, so we should be consistent with some other implementation here (maybe the airflow one) and also strive to be consistent with a default implementation everywhere. I'll open a ticket about that.

danielhoherd: Airflow has a pretty good securityContext design diagram in their sourcecode <https://github.com/apache/airflow/blob/9dd77520be3d8492156958d57b63b5779a3f55eb/chart/templates/_helpers.yaml#L848-L880>

pgvishnuram: those are strairght forward but we should mend our code in houston thats why we are going in selectives

danielhoherd (approved):
This looks good for now. It's not perfect, but it will unblock us. We can circle back to this when working on https://github.com/astronomer/issues/issues/6368

This is going to break older deployments though because it changes the `securityContexts` helm values structure.

Simpcyclassy:
Is this PR still needed?

pgvishnuram:
@Simpcyclassy yes this needed but requires a small code change at houston end too and we have not looked back this issue

danielhoherd (approved):
Looks good.

Shubham-astro:
@pgvishnuram @danielhoherd  Using this PR for https://github.com/astronomer/issues/issues/6729 as discussed, I have fixed a couple of issues can you review and let me know if there is anything that needs to be added please.?

danielhoherd (approved):
All the new changes look good to me.

PR #500, opened 2024-05-21
6233 dag rollback env

danielhoherd:
## Description

- Add `HOUSTON_SERVICE_ENDPOINT` env var
- Improve dag-server sts test cases

## Related Issues

- https://github.com/astronomer/issues/issues/6374
- https://github.com/astronomer/issues/issues/5623

## Testing

All of this code hasn't been in prod before and dag-deploy isn't going to work without it, so this can be tested during tests of dag-deploy rollback feature.

## Merging

Merge only into whatever release-0.35 is going to use.

### DIFF
--- a/templates/dag-deploy/dag-server-statefulset.yaml
+++ b/templates/dag-deploy/dag-server-statefulset.yaml
@@ -54,7 +54,7 @@ spec:
             {{- toYaml .Values.dagDeploy.resources | nindent 12 }}
           env:
           - name: HOUSTON_SERVICE_ENDPOINT
-            value: "https://{{ .Values.platform.release }}-houston.{{ .Values.platform.namespace }}.svc.cluster.local.:8871/v1/"
+            value: "http://{{ .Values.platform.release }}-houston.{{ .Values.platform.namespace }}.svc.cluster.local.:8871/v1/"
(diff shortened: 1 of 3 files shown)


templates/dag-deploy/dag-server-statefulset.yaml
```diff
@@ -53,6 +53,8 @@ spec:
           resources:
             {{- toYaml .Values.dagDeploy.resources | nindent 12 }}
           env:
+          - name: HOUSTON_SERVICE_ENDPOINT
+            value: "https://{{ .Values.platform.release }}-houston.{{ .Values.platform.namespace }}.svc.cluster.local.:8871/v1/"
```
pgvishnuram, line 57: https wont work and it will fail

rishkarajgi: @pgvishnuram can you elaborate?
`https wont work and it will fail`

danielhoherd: @rishkarajgi houston TLS is terminated at nginx when hit on the external URL. internally the houston service is only provided over HTTP.

rishkarajgi (approved):

PR #502, opened 2024-05-24
enhance nginx auth sidecar config

pgvishnuram:
## Description

This PR enhances existing authsidecar config and fixes code duplication.
change 1: client_max_body_size initially nested inside auth block its now moved to server block to get applied for all location

change 2: server_tokens is added  every where in auth-sidecar nginx block  now its moved to common location and called every to reduce code duplication

change 3:  localhost in upstream block connects ipv6 by default and falls back to ipv4 this was causing error in first try and fallbacks to ipv4 and call passed sucessfully - we have changed to 127.0.0.1 to connect ipv4 by default
 

## Related Issues

https://github.com/astronomer/issues/issues/6141

## Testing

QA should validate webserver, flower and dag-server to see services comes up without any issues.
This also now enhances our upload limit for all auth proxy components 

## Merging

cherry-pick to release-1.10, master and 1.11

### DIFF
--- a/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
+++ b/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
@@ -28,7 +28,7 @@ data:
       server {
         server_name {{ .Release.Name }}.{{ .Values.ingress.baseDomain }} ;
         listen {{ .Values.authSidecar.port }}  ;
-{{ include "default_nginx_server_settings" . | indent 8 }}
+        {{ include "default_nginx_server_settings" . }}
         location  = /auth {
 {{ include "default_nginx_settings" . | indent 8 }}
           proxy_set_header            Host                    houston.{{ .Values.ingress.baseDomain }};
--- a/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
+++ b/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
@@ -23,7 +23,7 @@ data:
       }
     http {
       upstream astro-dag-server {
-        server localhost:{{ .Values.dagDeploy.ports.dagServerHttp }} ;
+        server 127.0.0.1:{{ .Values.dagDeploy.ports.dagServerHttp }} ;
       }
       server {
         server_name {{ .Release.Name }}.{{ .Values.ingress.baseDomain }} ;
--- a/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
+++ b/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
@@ -28,7 +28,7 @@ data:
       server {
         server_name {{ .Release.Name }}.{{ .Values.ingress.baseDomain }} ;
         listen {{ .Values.authSidecar.port }}  ;
-        {{ include "default_nginx_server_settings" . }}
+{{ include "default_nginx_server_settings" . | indent 8 }}
         location  = /auth {
 {{ include "default_nginx_settings" . | indent 8 }}
           proxy_set_header            Host                    houston.{{ .Values.ingress.baseDomain }};
--- a/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
+++ b/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
@@ -28,10 +28,9 @@ data:
       server {
         server_name {{ .Release.Name }}.{{ .Values.ingress.baseDomain }} ;
         listen {{ .Values.authSidecar.port }}  ;
-        # Disable Nginx Server version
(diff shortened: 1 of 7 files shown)


templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
```diff
@@ -23,15 +23,14 @@ data:
       }
     http {
       upstream astro-dag-server {
-        server localhost:{{ .Values.dagDeploy.ports.dagServerHttp }} ;
+        server 127.0.0.1:{{ .Values.dagDeploy.ports.dagServerHttp }} ;
```
pgvishnuram, line 26: call ipv4 by default

templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
```diff
@@ -23,15 +23,14 @@ data:
...
       }
       server {
         server_name {{ .Release.Name }}.{{ .Values.ingress.baseDomain }} ;
         listen {{ .Values.authSidecar.port }}  ;
-        # Disable Nginx Server version
-        server_tokens off;
-{{ include "default_nginx_settings" . | indent 6 }}
+{{ include "default_nginx_server_settings" . | indent 8 }}
```
pgvishnuram, line 31: created new server location template to call them every known authsidecar block to reduce code duplication

rishkarajgi (approved):

danielhoherd (changes requested):
Looks good. Needs tests.

tests/chart/test_auth_sidecar.py
```diff
@@ -89,19 +90,26 @@ def test_auth_sidecar_config_with_dag_server_enabled(self, kube_version):
...
         assert volumeMounts in c_by_name["auth-proxy"]["volumeMounts"]
 
         assert docs[1]["kind"] == "Service"
         assert docs[1]["apiVersion"] == "v1"
         assert docs[1]["metadata"]["name"] == "release-name-dag-server"
+        print(docs[2]["data"]["nginx.conf"])
         assert authSidecarServicePorts in docs[1]["spec"]["ports"]
 
+        nginx_conf = pathlib.Path(
+            "tests/chart/test_data/dag-server-authsidecar-nginx.conf"
```
danielhoherd, line 109: If we're going to start doing this kind of thing, we'll need to come up with a way to organize these files better by naming them after the function that uses them, or by their content hash. We don't need to solve this now if we don't want to, but if we don't solve it now we'll have to solve it as soon as we start adding more files.

templates/webserver/webserver-auth-sidecar-configmap.yaml
```diff
@@ -28,8 +28,7 @@ data:
       server {
         server_name {{ .Release.Name }}.{{ .Values.ingress.baseDomain }} ;
         listen {{ .Values.authSidecar.port }}  ;
-        # Disable Nginx Server version
-        server_tokens off;
+{{ include "default_nginx_server_settings" . | indent 8 }}
```
danielhoherd, line 31: Since this isn't nindent of a data structure, should we just indent it so it aligns with the code, rather than calling `| indent 8` ?
```suggestion
        {{ include "default_nginx_server_settings" . }}
```

templates/flower/flower-auth-sidecar-configmap.yaml
```diff
@@ -28,8 +28,7 @@ data:
       server {
         server_name {{ .Release.Name }}-flower.{{ .Values.ingress.baseDomain }} ;
         listen {{ .Values.authSidecar.port }}  ;
-        # Disable Nginx Server version
-        server_tokens off;
+{{ include "default_nginx_server_settings" . | indent 8 }}
```
danielhoherd, line 31: Since this isn't nindent of a data structure, should we just indent it so it aligns with the code, rather than calling `| indent 8` ?
```suggestion
        {{ include "default_nginx_server_settings" . }}
```

templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
```diff
@@ -23,15 +23,14 @@ data:
...
         location  = /auth {
+{{ include "default_nginx_settings" . | indent 8 }}
```
danielhoherd, line 33: Since this isn't nindent of a data structure, should we just indent it so it aligns with the code, rather than calling `| indent 8` ?
```suggestion
        {{ include "default_nginx_server_settings" . }}
```

danielhoherd (approved):
Looks pretty good. I left some optional code suggestions, and also a note about thinking more about the test data naming convention.

rishkarajgi:
tests failing

PR #511, opened 2024-07-02
rename dag server ingress annotation to correct key

pgvishnuram:
## Description

rename dag server ingress annotation to correct key

## Related Issues

https://github.com/astronomer/issues/issues/6459

## Testing

No QA effort required

## Merging

cherry-pick to release-1.11


danielhoherd (changes requested):
Looks good, but it looks like we have no testing for this. We should take this opportunity to add a test for custom dagServerAnnotations values.

pgvishnuram:
we can implement jsonSchema for astronomer values as well for airflow chart that will help us atleast to make sure we dont pass some non supported values

danielhoherd:
@pgvishnuram in the past when we have done jsonSchema, we have not been diligent in maintaining it, which caused more problems than it solved, so we abandoned it. EG: last updated 2022-03-08 https://github.com/astronomer/astronomer/blob/master/values.schema.json.example

rishkarajgi:
How is this issue closed while the PR is still open? Is this a bug fix?

pgvishnuram:
since the new name changes is injected from houston and all other templates utilize the new naming - somehow this got identified at last minute while fixing reuse bug - it can go on next patch not a blocker

rishkarajgi (approved):

PR #514, opened 2024-07-11
Add dag-deploy networkPolicy

danielhoherd:
## Description

Add dag-deploy networkPolicy that prevents dag-deploy from being accessed from other namespaces except houston pod in astronomer namespace.
add support for BYO ingress with inbound auth sidecar 

## Related Issues

- <https://github.com/astronomer/issues/issues/6479>

## Testing

I manually tested this in an AWS cluster by confirming that when the policy was absent, dag-deploy server could be accessed from any namespace, and when this policy was in place, only the current namespace could access it, and also the houston pod in the astronomer deployed namespace.

## Merging

This should be merged into releases >= 0.34

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ test: unittest-chart ## Run all tests
 
 .PHONY: unittest-chart
 unittest-chart: charts venv ## Unittest the helm chart
-	# Protip: you can modify pytest behavior like: make unittest-chart PYTEST_ADDOPTS='-v --maxfail=1 --pdb -k 1.20'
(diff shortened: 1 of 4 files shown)


pgvishnuram:
@danielhoherd there is one more change need to be added

rishkarajgi:
> @danielhoherd there is one more change need to be added . we have one more option sidecar based ingress we need to include those changes as well

@pgvishnuram are you adding this change?

pgvishnuram:
> > @danielhoherd there is one more change need to be added . we have one more option sidecar based ingress we need to include those changes as well
> 
> @pgvishnuram is this change related to the same issue?

one change is done - other change im testing will added in another 30min

rishkarajgi (approved):

PR #517, opened 2024-07-23
add persistentVolumeClaimRetentionPolicy support in dag server

pgvishnuram:
## Description

Add persistentVolumeClaimRetentionPolicy support for dag server stateful to enable auto cleanup of pvc 

## Related Issues

https://github.com/astronomer/issues/issues/6436

## Testing

refer issue ticket

## Merging

cherry-pick to release 1.10, 1.11

### DIFF
--- a/templates/dag-deploy/dag-server-statefulset.yaml
+++ b/templates/dag-deploy/dag-server-statefulset.yaml
@@ -16,6 +16,9 @@ metadata:
       {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
+  {{- if  .Values.dagDeploy.persistentVolumeClaimRetentionPolicy }}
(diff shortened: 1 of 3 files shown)


pgvishnuram:
@danielhoherd shall i link this flag with persistance enabled true

templates/dag-deploy/dag-server-statefulset.yaml
```diff
@@ -2,6 +2,7 @@
 ## dag-server StatefulSet ##
 ############################
 {{- if .Values.dagDeploy.enabled }}
+{{- $persistenceEnabled := .Values.dagDeploy.persistence.enabled }}
```
danielhoherd, line 5: Why do we need this? Why can't we just use `.Values.dagDeploy.persistence.enabled` on line 20 below?

danielhoherd (approved):
Looks good. I left one question that I think leads to a simpler template, but it's definitely not a deal breaker.

PR #519, opened 2024-07-25
update exposed route paths for dag server

pgvishnuram:
## Description

update exposed route paths for dag server

PS: we already have a auth setup protecting these endpoints so we are safe here to expose them

## Related Issues

https://github.com/astronomer/issues/issues/6518

## Testing

QA should able to acess downloads and healthz endpoint 

## Merging

cherry-pick to release-1.11

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -193,7 +193,7 @@ spec:
           {{- if .Values.authSidecar.enabled  }}
           - path: /{{ .Release.Name }}/dags
           {{- else }}
-          - path: /{{ .Release.Name }}/dags/(upload|download)(/.*)?
+          - path: /{{ .Release.Name }}/dags/(upload|downloads|healthz)(/.*)?
           {{- end }}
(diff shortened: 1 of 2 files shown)


rishkarajgi (approved):

templates/ingress.yaml
```diff
@@ -193,7 +193,7 @@ spec:
           {{- if .Values.authSidecar.enabled  }}
           - path: /{{ .Release.Name }}/dags
           {{- else }}
-          - path: /{{ .Release.Name }}/dags/(upload|download)(/.*)?
+          - path: /{{ .Release.Name }}/dags/(upload|downloads|healthz)(/.*)?
```
danielhoherd, line 196: Is there any good reason to make downloads and healthz public?

himabindu07: @pgvishnuram It's easy for us to use this endpoint to check the dag-server validations in our automation

PR #523, opened 2024-08-07
add sidecar logs for dagserver

pgvishnuram:
## Description

Add sidecar logging support for dag server

## Related Issues

https://github.com/astronomer/issues/issues/6523
https://github.com/astronomer/issues/issues/6401

## Testing

QA should able to see logs in the astro ui

## Merging

cherry-pick to release-1.10, release-1.11


templates/dag-deploy/dag-server-statefulset.yaml
```diff
@@ -138,4 +175,4 @@ spec:
       {{- if .Values.dagDeploy.persistence.storageClass }}
       storageClassName: {{ .Values.dagDeploy.persistence.storageClass }}
       {{- end }}
-{{- end }}
+{{- end }}
```
Simpcyclassy, line 178: Do you need to add a new line?

danielhoherd (changes requested):
This all looks good so far. It doesn't have tests though, and when I went to look for tests of the vector sidecar I didn't find any. We should add tests for the vector sidecar containers. A test case that tests all pods where we should have a vector container would be great. AFAIK they should all be consistent, so testing them all with the same test case would be great.

templates/dag-deploy/dag-server-statefulset.yaml
```diff
@@ -114,14 +124,51 @@ spec:
...
+              fieldRef:
+                apiVersion: v1
+                fieldPath: metadata.labels['workspace']
+          volumeMounts:
+          - mountPath: /etc/sidecar-logging
+            name: config-volume
+          - name: sidecar-logging-consumer
+            mountPath: /var/log/sidecar-logging-consumer
+            readOnly: False
+{{- end }}
```
pgvishnuram, line 156: @danielhoherd we have this entire section similarity in git sync as well - if we render from helper templates we can reduce code repetition

danielhoherd: @pgvishnuram yeah, if we can reduce code duplication without complicating the templates then let's do it. This seems like something that should be identical everywhere.

danielhoherd (approved):
This looks fine. New tests are good, but I'd still like to see tests that cover the sidecar log consumer container across the whole chart. If that's outside the scope of this ticket then let's make a new ticket for that and merge this, otherwise we should add that test here.

PR #526, opened 2024-08-14
add airflow downgrade config to vector sidecar configuration list

pgvishnuram:
## Description

This PR adds support for airflow downgrade pod logs to logging server from our sidecar based mechanism

## Related Issues

https://github.com/astronomer/issues/issues/6401

## Testing

QA should able to view logs in astro-ui 

## Merging

cherry-pick to release-1.11


danielhoherd (approved):
Looks great. vector is so much better than fluentd. I think we should consider replacing our logging daemonset's fluentd with vector. https://github.com/astronomer/issues/issues/6474

PR #528, opened 2024-08-25
implement astro.dagDeploy.serviceAccount.create and astro.dagDeploy.service…

rob-1126:
## Description

Adds astro.dagDeploy.serviceAccount.create  and astro.dagDeploy.serviceAccount.name to let users specify a service account name and specify that they don't want the chart to create the service-account and associated rolebindings for the dag-deploy server.

## Related Issues

https://github.com/astronomer/issues/issues/6583

## Testing

Lightly tested by myself against 1.11.4.

## Merging

Cherry-pick to 1.11.x and main.


pgvishnuram:
This might require a Houston change

danielhoherd (changes requested):
Seems fine. Needs tests.

pgvishnuram:
@danielhoherd should the naming be in the format of release-name-airflow-dag-server or the existing format what we use  currently as  release-name-dag-server

values.yaml
```diff
@@ -564,6 +564,8 @@ dagDeploy:
 
   serviceAccount:
     annotations: {}
+    name: ""
```
danielhoherd, line 567: Shouldn't this be null?
```suggestion
    name: ~
```

danielhoherd:
> @danielhoherd should the naming be in the format of release-name-airflow-dag-server or the existing format what we use currently as release-name-dag-server

It's going to be in the airflow deployment namespace, so I think adding "airflow" is superfluous. Just like OSS airflow has removed "release-name" from their components because it is also superfluous.

So could we get away with "dag-server" ?

templates/_helpers.yaml
```diff
@@ -164,6 +164,18 @@ proxy_cookie_domain                     off;
 proxy_redirect                          off;
 {{ end }}
 
+{{/*
+Create the name of the dag-server service account to use
+*/}}
+{{- define "astro.dagDeploy.serviceAccountName" -}}
+{{- if .Values.dagDeploy.serviceAccount.create -}}
+    {{ default (printf "%s-dag-server" (include "airflow.fullname" .)) .Values.dagDeploy.serviceAccount.name }}
```
danielhoherd, line 172: Seems like we should have one more test assertion for each of these conditions. Currently we only have two assertions about `docs[0]["spec"]["template"]["spec"]["serviceAccountName"]`, but there are three possible code paths here.

danielhoherd (approved):
Looks good. I'd like to see one more test case though, which I left a comment about.

templates/_helpers.yaml
pgvishnuram: added missing checks as well

danielhoherd (approved):

PR #542, opened 2024-10-23
remove authsidecar tls secret hard codings

pgvishnuram:
## Description

remove auth sidecar tls secret name hard coding and sync with global tls secretname, this allows users to pass their own secret name that will synced across all deployments.

## Related Issues

https://github.com/astronomer/issues/issues/6344

## Testing

QA should able to pass custom tls secret name for authsidecar and regular deployment from astronomer global values

## Merging

cherry-pick to 1.11 and 1.13

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -96,13 +96,13 @@ metadata:
   {{- end }}
 spec:
   tls:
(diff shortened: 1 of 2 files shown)


Shubham-astro (approved):
Looks good

PR #544, opened 2024-10-24
update nginx worker connections and nginx worker process

pgvishnuram:
## Description

This PR fixes an issue where on baremetal installation worker process causing an OOM.

what is worker process ?
worker process that handles the incoming request
setting auto automatically adjust the number of Nginx worker processes based on available cores

what is worker connections ?
maximum number of simultaneous connections that can be opened by a worker process.

## Related Issues

- https://github.com/astronomer/issues/issues/6685

## Testing

QA should validate the values in nginx config when auth sidecar is enabled

## Merging

cherry-pick to all valid releases

### DIFF
--- a/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
+++ b/templates/dag-deploy/dag-server-auth-sidecar-configmap.yaml
@@ -15,10 +15,10 @@ metadata:
 data:
   nginx.conf: |
     pid /tmp/nginx.pid;
-    worker_processes auto;
+    worker_processes 1;
     events {
       multi_accept        on;
-      worker_connections  16384;
+      worker_connections  3072;
       use                 epoll;
       }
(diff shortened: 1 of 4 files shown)


pgvishnuram:
@danielhoherd @rob-1126 should we expose this as a user configurable option or let us hardcode to defaults

rujhan-arora-astronomer (commented):
Why did we chose 1 and 3072 specifically? @pgvishnuram Why not some other numbers?

pgvishnuram:
> Why did we chose 1 and 3072 specifically? @pgvishnuram Why not some other numbers?

1 is defaut and 3072 is 6 times the defaukt

danielhoherd (approved):
I have no idea if these numbers are wise, but they look like they won't break the code, so LGTM.

PR #558, opened 2025-01-23
add security context overrides support for kubernetes executor

pgvishnuram:
## Description

Add configurable security context overrides support pod templates

## Related Issues

- https://github.com/astronomer/issues/issues/6927

## Testing

QA should able to pass customer security context to kube pod executor task pods

## Merging

cherry-pick to release-1.13


values.yaml
```diff
@@ -201,6 +201,8 @@ airflow:
     {{- $nodeSelector := or .Values.nodeSelector .Values.workers.nodeSelector }}
     {{- $affinity := or .Values.affinity .Values.workers.affinity }}
     {{- $tolerations := or .Values.tolerations .Values.workers.tolerations }}
+    {{- $securityContext := include "airflowPodSecurityContext" (list . .Values.workers) }}
```
danielhoherd, line 204: Shouldn't this match the style of the line below and be:
```suggestion
    {{- $airflowPodSecurityContext := include "airflowPodSecurityContext" (list . .Values.workers) }}
```

danielhoherd (changes requested):
Seems like a good start. Needs tests.

values.yaml
danielhoherd: Where is this `airflowPodSecurityContext` template?

pgvishnuram: we are using airflow oss chart mappings here

pgvishnuram: yes if this naming suits we can rename it as long as we all know why is this being used here

Shubham-astro (approved):

PR #568, opened 2025-02-20
update airflow chart version 1.14.0

pgvishnuram:
## Description

update airflow chart version 1.14.0

## Related Issues

NA, ktlo work

## Testing

No testing required 

## Merging

cherry-pick to master and create new release brancg 1.14

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v2 is Helm 3
 apiVersion: v2
 name: airflow
-version: 1.13.6
+version: 1.14.0
 description: Helm chart to deploy the Astronomer Platform Airflow module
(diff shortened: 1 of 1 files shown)


Chart.yaml
```diff
@@ -1,7 +1,7 @@
 # apiVersion v2 is Helm 3
 apiVersion: v2
 name: airflow
-version: 1.13.6
+version: 1.14.0
```
danielhoherd, line 4: Why is this bump needed?

pgvishnuram: 0.36 and 0.37 will have different changes right or we are going put same features as before

danielhoherd: I see, so 1.14 is for 0.37. That makes sense.

danielhoherd (approved):

PR #571, opened 2025-02-28
Fixing Vector Sidecar logging Termination

Shubham-astro:
## Description

This PR addresses an issue with celery worker pods hanging during termination when sidecar-logging is enabled. When pods need to be deleted or replaced (such as during routine configuration changes), the termination process sometimes hangs until the full grace period (10 minutes) elapses, even when no celery worker tasks are running.

Root cause analysis revealed that the Vector logging sidecar container doesn't properly handle termination signals. While the main worker container receives SIGTERM and initiates shutdown procedures, the Vector sidecar continues to monitor log files indefinitely, causing the pod to remain in "Terminating" state until the full grace period expires.

This PR adds two key configuration parameters to the Vector sidecar configuration:
- `shutdown_timeout_secs: 10` - Ensures Vector exits within 10 seconds of receiving a termination signal
- `ignore_older_secs: 600` - Prevents Vector from processing files that haven't been modified in the last 10 minutes, helping it to properly complete its work during shutdown

## Related Issues

https://github.com/astronomer/issues/issues/6714

## Testing

Tested Manually.

- This is before (old worker pod is stuck in terminating state even after several minutes):
Screenshot 2025-02-28 at 1 56 50 PM

- This is after the change was done:
Screenshot 2025-02-28 at 12 48 10 PM

## Merging

Bug fix, This can be merged everywhere. 

## Risk Assessment

Low risk change that only affects the termination behavior of worker pods. The changes:
- Do not modify how logs are stored or accessed
- Do not affect normal operation of the logging pipeline
- Only impact behavior during pod termination
- Use standard Vector configuration parameters with conservative values


templates/logging-sidecar-configmap.yaml
```diff
@@ -23,6 +24,7 @@ data:
         include:
           - "${SIDECAR_LOGS}/*.log"
         read_from: beginning
+        ignore_older_secs: 600
```
danielhoherd, line 27: I'm curious why this is needed. Is this in case the sidecar logging container crashes and restarts?

templates/logging-sidecar-configmap.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   vector-config.yaml: |
+    shutdown_timeout_secs: 10
```
danielhoherd, line 17:
```
$ helm template . --show-only templates/logging-sidecar-configmap.yaml --set loggingSidecar.enabled=true --set airflow.elasticsearch.enabled=true --set airflow.elasticsearch.connection.host=example.com --set airflow.elasticsearch.connection.port=9999 --set airflow.elasticsearch.connection.user=THE_USER --set airflow.elasticsearch.connection.pass=THE_PASS | yq -r '.data.["vector-config.yaml"]' > vector.yaml
$ SIDECAR_LOGS=THE_SIDECAR_LOGS vector validate vector.yaml
Failed to load ["vector.yaml"]
------------------------------
x unknown field `shutdown_timeout_secs`
```

danielhoherd: Ooh, I bet we could easily make a pre-commit hook to do that validation.

pgvishnuram: https://vector.dev/docs/reference/cli/#vector_graceful_shutdown_limit_secs

Shubham-astro: I thought shutdown_timeout_secs was also a global config that can be put in the configmap, but looks like it isn't. It seems to be present under components like: https://vector.dev/docs/reference/configuration/sources/socket/#shutdown_timeout_secs

templates/logging-sidecar-configmap.yaml
Shubham-astro: This is not needed as per the case but I added it as an improvement that the vector doesn't look at files which have not had any movement in last 10 mins. We can discuss if this is not needed.

templates/logging-sidecar-configmap.yaml
Shubham-astro: Although I don't think this is going to work as the documentation says the default value says default value is 60s and from ticket and our testing this seems to be running for around 10 mins.

PR #592, opened 2025-05-30
add base init job for git sync repo shared mode

pgvishnuram:
## Description

Adds a `pre-install`/`pre-upgrade` Helm hook Job for `gitSyncRelay.repoShareMode: shared_volume` that clones the configured repo into the `git-repo-contents` PVC before Helm creates/upgrades any of the Airflow component Deployments that mount it. Without this, an Airflow component pod (e.g. the scheduler) can mount the `dags` subPath before git-sync-relay's own Deployment has done its first clone, producing "Stale file handle" (ESTALE) on that mount — this closes that race for the shared_volume repo-share mode.

**Update:** extended to support HTTPS+PAT auth (the hook originally only wired SSH, predating HTTPS+PAT support). The hook Job now uses the `ap-git-sync-relay` image and its one-shot `git-sync-once` entrypoint (astronomer/git-sync-relay#71) instead of the upstream `git-sync` image — upstream `git-sync` has no equivalent of git-sync-relay's in-process HTTPS+PAT credential helper, and wiring PAT against its own auth mechanism would put the token in a plaintext env var. HTTPS+PAT is now wired the same way the long-running Deployment does it (`git-https-secret` volume, `GIT_SYNC_AUTH_TYPE`, `GIT_SYNC_HTTPS_SECRET_DIR`). The auth-type/HTTPS-secret resolution and validation logic (previously inline only in `git-sync-relay-deployment.yaml`) is now shared between the Deployment and this hook Job via new `_helpers.yaml` templates, so the two can't validate this differently. Also fixed a securityContext gap (the container only had `readOnlyRootFilesystem` set, missing the capability-dropping the Deployment already has) and added `activeDeadlineSeconds` (from `repo.gitSyncTimeout`) since a one-shot Job has no in-process retry loop to bound its own runtime the way the Deployment does.

Known gap not addressed here: this hook Job has no private-CA trust wiring (`global.privateCaCerts`), unlike the Deployment. Pre-existing in the original draft, not introduced by the HTTPS+PAT extension — a customer using both `global.privateCaCerts` and `shared_volume` mode would need this added separately.

## Related Issues

- https://github.com/astronomer/issues/issues/7262
- [PINF-1115](https://linear.app/astronomer/issue/PINF-1115) — Shared Volume "Stale file handle" on the dags mount (this PR is the fix for the reported race)
- [PINF-1190](https://linear.app/astronomer/issue/PINF-1190) — this PR's own tracking ticket for the HTTPS+PAT extension
- [PINF-1188](https://linear.app/astronomer/issue/PINF-1188) — git-sync-relay's `git-sync-once` entrypoint this hook now depends on

## Testing

- Unit tests: `tests/chart/test_git_sync_relay_init_job.py` updated (image, `git-sync-once` invocation, securityContext, `activeDeadlineSeconds`) and extended (HTTPS+PAT, https-none, SSH+HTTPS mutual exclusivity, invalid auth type)
- `tests/chart/test_git_sync_relay_deployment.py` and `test_git_sync_relay_pvc.py` re-run to confirm the shared-helper refactor is render-identical for the Deployment (155 + rest passed, unchanged)
- Full chart suite: 1210 passed, 46 skipped (pre-existing), 0 failed
- `helm lint` clean
- **Still pending cluster test** (as originally noted) — this PR has never been validated against a real cluster, and this depends on [PINF-1188](https://linear.app/astronomer/issue/PINF-1188) shipping a released `ap-git-sync-relay` image tag with `git-sync-once` before the HTTPS+PAT path is functionally live, not just chart-correct

## Merging

Master, 1.17

## Critical review round (2026-08-14)

A critical self-review against this project's design history (private-CA trust exists specifically for enterprise Git hosts, per the archived HTTPS+PAT design docs) surfaced two fixable gaps, now fixed:

- **Private-CA trust was missing entirely from the init Job**, unlike the Deployment. A `shared_volume` deployment against a private-CA git host would have its bootstrap clone fail TLS validation even though the long-running Deployment (which has this wiring) would work fine once started. Wired in the same `gitSyncRelay.customCa*` helpers the Deployment already uses.
- **No `backoffLimit`** — Kubernetes' default of 6 would silently retry a failed clone, and since `activeDeadlineSeconds` bounds the Job's *total* time across all attempts (not per-attempt), it would end up bounding a fraction of `gitSyncTimeout` per attempt instead of the full value. Added `backoffLimit: 0` so a failure surfaces immediately.

Three more findings from that review needed live-cluster verification or further work, not more render tests, at the time:
1. ~~The PVC hook is `pre-install`-only while the Job is `pre-install,pre-upgrade`...~~ **Fixed** in a later round (see below) — the PVC's own render is now `lookup`-gated and fires on `pre-upgrade` too.
2. This hook re-runs on every `helm upgrade`, doing a real clone/fetch/reset against the same PVC the long-running Deployment may be concurrently mid-sync on (via its own poll/webhook cycle) — no coordination between the two processes today. **Still open, tracked as PINF-1194**; documented directly in the chart now (see the latest round below), not just in a scenario test docstring.
3. Neither the Deployment's nor the Job's `git-config-manager` init container sets `HOME` explicitly — both assume the `ap-git-sync-relay` image defaults `HOME=/home/git-sync` internally. If wrong, `safe.directory` silently never applies. **Still open**, not yet investigated.

## Chart-default bump (2026-08-14)

Bumped `gitSyncRelay.images.gitSync` default tag from `0.2.1` to `0.5.0` (the version [git-sync-relay#71](https://github.com/astronomer/git-sync-relay/pull/71) just merged as, which includes `git-sync-once`). Verified `0.5.0` is actually published before bumping.

Worth being explicit about what this does and doesn't do: real deployments don't run this chart default at all — houston-api generates each Airflow Deployment's Helm values, and the actual image tag comes from the `astronomer` repo's own `values.yaml` override (`global.gitSyncRelay.images.gitSync.tag`, currently `0.4.3` in prod), not this file. This bump alone doesn't reach real deployments; that needs the `astronomer` repo's own override repinned separately, same propagation chain as `airflowChartVersion`.

292 passed across the three related test files, `helm lint` clean.

## Human review round (2026-08-21)

rishkarajgi's review (advisory REQUEST CHANGES) caught a real blocker the render-test suite couldn't see:

- **Blocker, fixed**: on `origin/master` the `git-repo-contents` PVC is a plain, non-hook resource. This PR had moved its entire definition into the init-job's hook block, which meant an existing `shared_volume` install's PVC would disappear from the normal-resource manifest on its first upgrade to this chart version and get pruned by Helm — wedging the upgrade and losing the cloned repo. Restored `git-sync-relay-pvc.yaml` to render the identical plain resource, coexisting with the hook block (which still solves the real ordering problem: Helm applies pre-install hooks before any normal resource, so the PVC can't live *only* as a normal resource without deadlocking a fresh install against the Job). Same fix resolves the should-fix about `helm uninstall` orphaning the PVC — it's a normal resource again, so uninstall cleans it up exactly as it always did.
- Documented the concurrent-writer-race (PINF-1194, still open) and the availability-coupling (pre-upgrade gates every upgrade on this Job) should-fixes directly on the Job's own chart annotations, not just in a test docstring.
- Nit: aligned the Job's `chart` label on `{{ .Chart.Name }}`, matching the PVC/ServiceAccount in the same file.

Two earlier Copilot comments (missing scheduling fields on the Job's pod spec; PVC not created on a mode-switch upgrade) were already fixed in a prior commit before this review round landed.

New render-test coverage in `test_git_sync_relay_pvc.py` for the restored plain resource. Full chart suite: 1252 passed, 46 skipped (pre-existing). Verified with a full `helm template` render (not just `show_only`-scoped tests) that both PVC declarations coexist cleanly with identical specs.


templates/git-sync-relay/git-sync-relay-init-job.yaml
```diff
@@ -0,0 +1,186 @@
...
+      labels:
+        component: git-sync-relay
+        tier: airflow
+        release: {{ .Release.Name }}
+    spec:
+      {{- if or .Values.airflow.registry.secretName .Values.airflow.registry.connection }}
+      imagePullSecrets:
+        - name: {{ template "astro.registry_secret" . }}
+      {{- end }}
+      restartPolicy: Never
```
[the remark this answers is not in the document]
danielhoherd: Fixed in d88a05ef — mirrored the Deployment's `$nodeSelector`/`$affinity`/`$tolerations` computation and rendered nodeSelector/affinity/tolerations/terminationGracePeriodSeconds on the Job's pod spec, same as the Deployment. New tests cover both airflow-level and gitSyncRelay-level precedence.

templates/git-sync-relay/git-sync-relay-init-job.yaml
```diff
@@ -0,0 +1,186 @@
...
+{{- $httpsSecret := include "gitSyncRelay.httpsSecretName" . }}
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: git-repo-contents
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+    "helm.sh/hook-weight": "1"
```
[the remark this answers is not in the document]
danielhoherd: Fixed in d88a05ef. Confirmed independently in my own critical review before this comment landed (same finding). Gated the PVC's render on `lookup "v1" "PersistentVolumeClaim" .Release.Namespace "git-repo-contents"` and added pre-upgrade to its hook, rather than just adding pre-upgrade on its own -- the latter would have made the existing before-hook-creation delete-policy destroy and recreate the PVC on *every* upgrade (not just a mode-switch one), reopening the exact ESTALE race this chart exists to close for any pod that already has it mounted. The lookup gate means the block only renders when the PVC doesn't already exist: unchanged on a normal upgrade, and now also created on an upgrade that newly enables shared_volume mode. `lookup` always returns empty during `helm template`/render tests (no live cluster), so this doesn't affect existing test coverage -- new test added for the updated hook annotation, and a live-cluster scenario test (in progress, separate PR in the astronomer repo) will exercise the actual install-then-upgrade transition.

danielhoherd:
The 5 "suppressed" private-CA comments on the latest review (customCaVolumes/customCaInitContainer/customCaVolumeMounts/customCaEnv/OpenShift-podSecurityContext-for-securityContext) are all stale — that review ran against `05c8c2bc`, before `a72aac0d` (private-CA wiring) and `d88a05ef` (this round, including the OpenShift podSecurityContext fix) landed. All 5 are addressed as of `d88a05ef`.

templates/git-sync-relay/git-sync-relay-init-job.yaml
```diff
@@ -0,0 +1,215 @@
...
+       delete and recreate the PVC -- destroying the cloned repo, and briefly reopening
+       the exact ESTALE race this chart exists to close -- on every single upgrade, not
+       just the first install. `lookup` only queries the live cluster (never during
+       `helm template`/render tests, where it always returns an empty dict), so this is
+       a no-op for the existing render-test coverage and only matters for a real
+       install/upgrade: a fresh install renders and creates it as before; an upgrade
+       where it already exists skips this block entirely (no hook, no risk to existing
+       data); an upgrade that newly enables shared_volume mode (no PVC yet) now creates
+       it via the pre-upgrade hook instead of failing the Job that expects to mount it. */}}
+{{- if not (lookup "v1" "PersistentVolumeClaim" .Release.Namespace "git-repo-contents") }}
```
rishkarajgi, line 24: **Conditional Blocker — verify before merge.** This gate correctly stops per-upgrade re-creation churn, but it doesn't address the one-time *normal→hook* manifest prune. On `origin/master` the PVC was a normal, release-tracked resource; this PR moves it entirely into this hook. On the first upgrade of an existing `shared_volume` install, the PVC is in the old manifest but absent from the new one (normal template removed; this hook skipped because the PVC already exists) → Helm 3 prunes it, PVC-protection holds it `Terminating` while mounted, and the upgrade wedges / loses cloned data. Keep the PVC a normal resource with `helm.sh/resource-policy: keep`, or add an explicit adoption step. (Only bites if field `shared_volume` installs exist — if pre-GA with none, this is a migration-note Should-fix.)

templates/git-sync-relay/git-sync-relay-init-job.yaml
```diff
@@ -0,0 +1,215 @@
...
+        - name: git-sync
+          image: "{{ .Values.gitSyncRelay.images.gitSync.repository }}:{{ .Values.gitSyncRelay.images.gitSync.tag }}"
+          {{- /* No `command` override: let the image's own entrypoint.sh run first (so
+                 CA-trust/debug setup still applies), then exec this instead of the
+                 default long-running Sanic daemon. Reuses the ap-git-sync-relay image
+                 and its credential-helper wiring (SSH / HTTPS+PAT) instead of the
+                 upstream git-sync image, which has no equivalent for HTTPS+PAT --
+                 wiring PAT auth against its native mechanism would put the token in a
+                 plaintext env var. */}}
+          args: ["git-sync-once"]
```
rishkarajgi, line 159: Should-fix (upgrade race): on `helm upgrade` this clones into `/git` while the prior-revision git-sync-relay Deployment is still Running and fetching the same PVC (hooks run before the normal-resource upgrade). Concurrent git operations can corrupt the worktree or collide on `.git/index.lock`; with `backoffLimit: 0` that aborts the whole upgrade. Appears unexercised (PINF-1194) — consider serializing against the relay or documenting the constraint.

templates/git-sync-relay/git-sync-relay-init-job.yaml
```diff
@@ -0,0 +1,215 @@
...
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-git-sync-relay-init
+  labels:
+    component: git-sync-relay
+    tier: airflow
+    release: {{ .Release.Name }}
+    chart: "{{ .Release.Name }}-{{ .Chart.Version }}"
```
rishkarajgi: Nit: this `chart` label uses `{{ .Release.Name }}`, but the hook PVC and ServiceAccount in this same file use `{{ .Chart.Name }}`. Align on `{{ .Chart.Name }}` for consistency (the base PVC had the same quirk).

rishkarajgi (commented):
**Verdict: REQUEST CHANGES (advisory)** — 1 conditional Blocker on the upgrade path, plus a few should-fixes dominated by upgrade/idempotency. The core fix (pre-populate the PVC before components mount it) is well-built — I want to flag the upgrade-migration edge before it ships.

**Credited as correct (verified against `origin/master`):** the whole init-Job path renders only under `gitSyncRelay.enabled && repoShareMode == shared_volume` (default is `git_daemon`, so other modes are untouched); the PVC is `ReadWriteMany` (no RWO scheduling trap); hook ordering (PVC weight 1 → SA 3 → Job 5) makes Helm wait for the clone before applying the consumers; **step-failure policy is explicit and fail-closed** (`backoffLimit: 0` + `activeDeadlineSeconds`), which is the right call; and there's strong render-test coverage.

**⚠️ Blocker (conditional — please verify): the PVC changes from a normal resource to a Helm hook, which can make Helm *prune* it on the first upgrade of an existing `shared_volume` install → data loss.**
Verified: on `origin/master`, `git-sync-relay-pvc.yaml` defines a normal, release-tracked PVC (no hook annotation, no `resource-policy`). This PR deletes that definition and re-creates the PVC as a `pre-install,pre-upgrade` hook inside the init-job, `lookup`-gated so it doesn't render when the PVC already exists. The hazard: for an install that already has the PVC, on upgrade it's in the *old* manifest but absent from the *new* one (normal template gone; hook copy skipped by the lookup gate) — Helm 3 prunes resources dropped from the manifest, and PVC-protection then holds it `Terminating` while mounted, wedging the upgrade / losing the cloned data. The in-file comment states "It persists across upgrades" — that's the assumption worth confirming, because the normal→hook transition may not deliver it. Fix direction: keep the PVC a normal resource with `helm.sh/resource-policy: keep`, or add an explicit adoption/migration step. **Caveat:** this only bites if `shared_volume` installs exist on the prior chart — if it's pre-GA with zero field installs, it's a Should-fix (migration note), not a Blocker. (Labeled Inference, not asserted Fact — worth a quick `helm upgrade` test on a pre-populated PVC.)

**Should-fix:**
1. **Concurrent-writer race on upgrade:** the gating hook clones into `/git` while the prior-revision relay Deployment is still Running and fetching the same PVC (hooks run before the normal-resource upgrade). Concurrent git ops can corrupt the worktree / collide on `.git/index.lock`, and with `backoffLimit: 0` that aborts the whole upgrade. Ships unexercised (the scenario docstring acknowledges PINF-1194) — serialize against the relay or document the risk.
2. **Availability coupling:** `pre-upgrade` fires on *every* `helm upgrade`, so a transient git-host outage or a clone exceeding `activeDeadlineSeconds` (300s default) blocks unrelated deployment upgrades. Intended fail-safe — worth stating explicitly with the rollback story.
3. **Uninstall leak:** hook resources aren't release-tracked, so `git-repo-contents` + its PV orphan on deployment deletion (previously cleaned up). Same root cause as the Blocker.

**Nits:** the Job's `chart` label uses `{{ .Release.Name }}` where the hook PVC/SA in the same file use `{{ .Chart.Name }}` (align them); a test docstring references a non-existent `test_init_job_pvc_lookup_gate_comment`.

templates/git-sync-relay/git-sync-relay-init-job.yaml
danielhoherd: Confirmed against `origin/master` -- the PVC there is a plain, non-hook resource, so this is real, not hypothetical. Fixed by restoring `git-sync-relay-pvc.yaml` to render the same plain resource (unchanged spec, so Helm's normal-resource apply is a no-op against whatever the hook already created), while keeping this hook block as-is for the actual gap it solves (ordering on first-time creation). The plain declaration keeps the PVC permanently tracked as a normal release resource, so it's never at risk of the normal→hook manifest-prune you flagged. Verified with a full `helm template` render (not just the show_only-scoped tests) that both declarations coexist cleanly with identical specs.

templates/git-sync-relay/git-sync-relay-init-job.yaml
danielhoherd: Fair -- serializing against the relay is a bigger design change than fits a review response, and it's already tracked separately (PINF-1194). Added an explicit comment on the Job's own annotations block documenting this as a known, unserialized gap rather than leaving it only in the astronomer scenario's test docstring where you'd have had to go looking for it.

templates/git-sync-relay/git-sync-relay-init-job.yaml
danielhoherd: Fixed -- aligned on {{ .Chart.Name }} (also fixed the same quirk in the restored plain PVC resource while I was there).

danielhoherd:
Addressed the review from rishkarajgi (replied inline on the Blocker + should-fixes/nit) and confirmed the two earlier Copilot comments (scheduling fields, mode-switch PVC) were already fixed in d88a05ef before this round landed.

**Blocker (PVC normal→hook migration prune risk) — fixed.** Confirmed real against `origin/master`: the PVC there is a plain, non-hook resource. This PR had moved its entire definition into the hook block, which meant an existing shared_volume install's PVC would disappear from the normal-resource manifest on upgrade and get pruned by Helm. Restored `git-sync-relay-pvc.yaml` to render the identical plain resource — it now coexists with the hook block, which still solves the one real problem it needs to (guaranteeing the PVC exists before the pre-install/pre-upgrade Job mounts it, since Helm applies pre-install hooks before any normal resource). Helm's normal-resource apply against an already-hook-created PVC is a no-op (same spec), and the plain declaration keeps it permanently tracked so it's never at risk of the prune again. Verified with a full `helm template` render, not just the show_only-scoped tests, that both coexist cleanly. New render-test coverage in `test_git_sync_relay_pvc.py` for the restored plain resource.

This same fix resolves the **uninstall-leak should-fix**: since the PVC is a normal resource again (matching `origin/master`), `helm uninstall` cleans it up exactly as it always did — no orphan.

**Should-fix (availability coupling)** — documented explicitly on the Job's own annotations block rather than only in the astronomer scenario's test docstring: a git-host outage or a clone exceeding `activeDeadlineSeconds` blocks that upgrade (intended fail-safe), and Helm doesn't auto-rollback hook failures by default — callers wanting that need `helm upgrade --atomic` themselves.

**Should-fix (concurrent-writer race)** — still deferred to PINF-1194 as before (a real fix here is a bigger design change than fits a review round), now also documented in-chart, not just in the test docstring.

**Nit (chart label consistency)** — fixed, aligned on `{{ .Chart.Name }}` everywhere in this file, including the restored plain PVC.

Full chart suite: `uv run pytest tests/chart/` — 1252 passed, 46 skipped (pre-existing).

PR #611, opened 2025-07-11
Sync 1 15 with master

danielhoherd:
## Description

Somehow master branch is out of sync with 1.15. This puts them back in sync.

## Related Issues

Discovered while working on https://github.com/astronomer/issues/issues/7452

## Testing

We just need to make sure this merge creates the right content in master branch.

## Merging

Merge only to master and then make sure master and 1.15 are in sync.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ jobs:
   run_pre_commit:
     resource_class: small
     docker:
-      - image: quay.io/astronomer/ci-pre-commit:2025-05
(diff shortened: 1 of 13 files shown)


danielhoherd:
This PR was merged to master. I merged it to release-1.15, verified that they both had the same contents (but not the same commit history), and then recreated the release-1.15 branch from the tip of master. This makes release-1.15 and master identical as of commit 5467e22121d8da651f67106adf02c00743754739

PR #613, opened 2025-08-05
add taint, toleration and node selector support

pgvishnuram:
## Description

add taint,tolleration and node selector support for git-sync-relay and dag-server

## Related Issues

- https://github.com/astronomer/issues/issues/7526
- https://github.com/astronomer/issues/issues/7335

## Testing

QA should able to override taint,tolleration and node selector for git-sync and dag-server components

## Merging

merge to master and release-1.15

### DIFF
--- a/tests/chart/conftest.py
+++ b/tests/chart/conftest.py
@@ -73,3 +73,34 @@ def docker_client():
         client = docker.from_env()
         yield client
         client.close()
+
+
+@pytest.fixture(scope="function")
+def airflow_node_pool_config():
+    yield {
+        "nodeSelector": {"role": "airflow"},
+        "affinity": {
+            "nodeAffinity": {
+                "requiredDuringSchedulingIgnoredDuringExecution": {
+                    "nodeSelectorTerms": [
+                        {
+                            "matchExpressions": [
+                                {
+                                    "key": "astronomer.io/multi-tenant",
+                                    "operator": "In",
+                                    "values": ["false"],
+                                }
+                            ]
+                        }
+                    ]
+                }
+            }
+        },
+        "tolerations": [
+            {
+                "effect": "NoSchedule",
+                "key": "airflow",
+                "operator": "Exists",
+            }
+        ],
+    }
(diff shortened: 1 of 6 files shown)


Shubham-astro (approved):
Looks good

tests/chart/test_dag_server_statefulset.py
```diff
@@ -318,3 +322,27 @@ def test_dag_server_service_account_with_template(self, kube_version):
         doc = docs[0]
         assert doc["kind"] == "StatefulSet"
         assert doc["metadata"]["name"] == "release-name-dag-server"
+
+    def test_dag_server_affinity(self, kube_version, airflow_node_pool_config):
+        """Test that dagserver affinity correctly inserts affinity, node pool and tolleration configs."""
```
danielhoherd, line 327:
```suggestion
        """Test that dagserver affinity correctly inserts affinity, node pool and toleration configs."""
```

tests/chart/test_git_sync_relay_deployment.py
```diff
@@ -368,3 +372,25 @@ def test_git_sync_service_account_with_template(self, kube_version):
         doc = docs[0]
         assert doc["kind"] == "Deployment"
         assert doc["metadata"]["name"] == "release-name-git-sync-relay"
+
+    def test_git_sync_relay_affinity(self, kube_version, airflow_node_pool_config):
+        """Test that git sync relay affinity, node pool and tolleration configs."""
```
danielhoherd, line 377:
```suggestion
        """Test that git sync relay affinity, node pool and toleration configs."""
```

danielhoherd (approved):
The changes are mostly good. I'm a bit concerned about this new fixture we're introducing that violates Locality of Behavior and creates a new abstraction that does not map 1:1 onto the k8s concepts. I don't want that to block shipping the feature though, especially since it only relates to test code. It would be nice if we could come up with a better solution now, but if we can't do so quickly then we can just merge this and circle back on it if it becomes a problem.

tests/chart/conftest.py
```diff
@@ -73,3 +73,34 @@ def docker_client():
...
+            }
+        },
+        "tolerations": [
+            {
+                "effect": "NoSchedule",
+                "key": "airflow",
+                "operator": "Exists",
+            }
+        ],
+    }
```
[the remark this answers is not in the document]
pgvishnuram: @danielhoherd this is the same logic from our astronomer repo

PR #615, opened 2025-08-11
fix workers scheduling ordering

pgvishnuram:
## Description

currently airflow toleration, nodeselector and affinity takes higher precedence than worker specific configuration which is incorrect with this new change worker configuration takes higher priority than airflow defaults

## Related Issues

- https://github.com/astronomer/issues/issues/7615

## Testing

QA should able to pass customer nodeSelector, tolerations and affinity config to workers section to see those values are getting higher precedence order than defaults

## Merging

merge to master, release-1.15


rishkarajgi (dismissed):
please add/update tests

PR #616, opened 2025-08-11
airflow oss helm chart 1.17.x official release changes

pgvishnuram:
## Description

Enables OSS helm chart 1.17 with airflow3 support

## Related Issues

- https://github.com/astronomer/issues/issues/7388

## Testing

QA should able to create airflow2 and airflow3 based deployments from astronomer platform

## Merging

merge to master

### DIFF
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -90,6 +90,15 @@
         key: connection
     {{- end }}
   {{- end }}
+  {{- if semverCompare ">=3.0.0" .Values.airflowVersion }}
+  - name: AIRFLOW__API_AUTH__JWT_SECRET
+    valueFrom:
+      secretKeyRef:
+        name: {{ template "fernet_key_secret" . }}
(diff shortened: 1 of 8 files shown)


templates/_helpers.yaml
```diff
@@ -90,6 +90,15 @@
         key: connection
     {{- end }}
   {{- end }}
+  {{- if semverCompare ">=3.0.0" .Values.airflowVersion }}
```
danielhoherd, line 93: Is this going to work? We need to make sure this is comparing the **airflow** version since runtime does not use semver. As long as `airflowVersion` is actually airflow and not runtime then this should be ok.

danielhoherd (approved):

templates/_helpers.yaml
pgvishnuram: we are using this chart for our current testing and works as expected since we pass the actual airflow version in this fields

PR #629, opened 2025-10-02
fix dod and git sync network policies

pgvishnuram:
## Description

fix dod and git sync network policies

## Related Issues

Yet to update

## Testing

Yet to update

## Merging

Yet to update

### DIFF
--- a/templates/dag-deploy/dag-deploy-networkpolicy.yaml
+++ b/templates/dag-deploy/dag-deploy-networkpolicy.yaml
@@ -55,7 +55,7 @@ spec:
           "kubernetes.io/metadata.name": {{ .Values.platform.namespace }}
       podSelector:
         matchLabels:
-          component: ingress-controller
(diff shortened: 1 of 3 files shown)


tests/chart/test_dag_deploy_networkpolicy.py
```diff
@@ -30,6 +30,34 @@ def test_gsr_networkpolicy_defaults_with_authsidecar_enabled(self, kube_version)
         )
         assert len(docs) == 0
 
+    def test_dag_deploy_networkpolicy_dag_deploy_enabled_with_dataplane_mode(self, kube_version):
+        """Test that a valid networkPolicy are rendered when dag-deploy is enabled."""
```
danielhoherd, line 34: We should parametrize this test for all three modes. Also need to fix the docstring:
```suggestion
    def test_dag_deploy_networkpolicy_dag_deploy_enabled_with_dataplane_mode(self, kube_version):
        """Test that a valid networkPolicy is rendered when dag-deploy and networkPolicies are enabled."""
```

PR #695, opened 2026-02-12
PINF-279 : add extra volume and volume mounts support

pgvishnuram:
## Description

PINF-279 - add extra volume and volume mounts support

## Related Issues

PINF-279 - https://linear.app/astronomer/issue/PINF-279/adds-support-for-privatecacerts-in-dag-server

## Testing

QA should able to add extraVolume and volumeMounts to dag server components

## Merging

merge to master and release-1.0


danielhoherd (dismissed):
Needs tests. These could go into a general test that sets a variety of configurations and tests them, and doesn't necessarily need its own test case.

Shubham-astro (approved):
simple addition of extravolume and volumemounts looks good.

PR #722, opened 2026-04-14
PINF-492: Make TLS configuration conditional across all ingress

Shubham-astro:
## Description

This PR makes TLS configuration conditional across all ingress templates so the tls block is only rendered when ingress.tlsSecretName is explicitly set, preventing invalid ingress specs with secretName: ~ 

## Related Issues

https://linear.app/astronomer/issue/PINF-492/conditionally-render-tls-block-in-ingress-templates-when-tlssecret-is

## Testing

- Unittests updated.

## Merging

Master, 1.17

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -33,7 +33,7 @@ metadata:
 spec:
   {{- if .Values.ingress.tlsSecretName }}
   tls:
-    - secretName: {{ .Values.ingress.tlsSecretName | default "~" }}
+    - secretName: {{ .Values.ingress.tlsSecretName }}
       hosts:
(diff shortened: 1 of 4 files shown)


templates/ingress.yaml
```diff
@@ -31,11 +31,13 @@ metadata:
   {{- end }}
   {{- end }}
 spec:
+  {{- if .Values.ingress.tlsSecretName }}
   tls:
     - secretName: {{ .Values.ingress.tlsSecretName | default "~" }}
```
pgvishnuram: do we still need to have default to ~ should no longer be needed

Shubham-astro: removed.

pgvishnuram (approved):

PR #726, opened 2026-05-06
PINF-559: Strip fsGroup and runAsUser from git-sync-relay pod securityContext on OpenShift

Shubham-astro:
## Description

On OpenShift clusters, git-sync-relay pods fail to schedule with an SCC violation because the pod-level securityContext has fsGroup: 65533 and runAsUser: 50000 hardcoded, which fall outside the namespace-allocated UID range (e.g. [1000780000, 1000789999]).
Passing securityContext: null from the Houston ConfigMap was not sufficient because customers could still set runAsUser/fsGroup directly in their deployment values, causing the same breakage.

## Related Issues

https://linear.app/astronomer/issue/PINF-559/git-sync-merging-security-context-with-default-values

## Testing
- Tested this on a cluster:
Screenshot 2026-05-13 at 6 35 27 PM

## Merging

Master, 1.17


values.yaml
```diff
@@ -655,6 +655,9 @@ airflow:
 # Enable security context constraints required for OpenShift
 sccEnabled: false
 
+openshift:
+  enabled: false
```
rishkarajgi, line 659: There's already a `sccEnabled: false` a couple lines up (line 656) that seems to serve a similar purpose. Using a separate `openshift.enabled` key risks these two flags drifting out of sync — e.g. someone sets `sccEnabled: true` for other SCC-related logic but forgets to also set `openshift.enabled: true`. Is there a reason not to reuse `sccEnabled` here, or should we document the distinction between the two?

templates/_helpers.yaml
```diff
@@ -479,3 +479,16 @@ proxy_buffering on;
...
+
+{{/*
+git-sync-relay pod-level securityContext
+Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
+*/}}
+{{- define "gitSyncRelay.podSecurityContext" -}}
+{{- if .Values.openshift.enabled }}
+{{- $required := dict "runAsNonRoot" true }}
+{{- merge $required (omit (.Values.gitSyncRelay.securityContext | default dict) "runAsUser" "fsGroup") | toYaml }}
+{{- else }}
```
rishkarajgi, line 491: Helm's `merge` gives priority to the first dict, so `runAsNonRoot: true` from `$required` will always win — even if a user explicitly sets `runAsNonRoot: false` in their values. That seems intentional (you always want non-root on OpenShift), but worth a comment in the template so it's not surprising to whoever reads this later.

rishkarajgi (commented):
Fixes SCC violations on OpenShift by stripping `fsGroup` and `runAsUser` from the git-sync-relay pod securityContext when `openshift.enabled` is true, and enforcing `runAsNonRoot: true` in their place. The approach is clean and the two new tests cover both the OpenShift and non-OpenShift paths well.

**Verdict:** 💬 A few questions — the `sccEnabled` vs `openshift.enabled` duplication and the `merge` priority behavior are worth confirming before this goes in.

values.yaml
pgvishnuram: sccEnabled is different and openshiftEnabled is different both service different purposes

pgvishnuram: we are not recommending users to use sccEnabled and any new installations or ongoing installations are recommended to use openshift.enabled to avoid security risks

pgvishnuram (approved):

PR #727, opened 2026-05-13
PINF-559: Strip fsGroup and runAsUser from git-sync-relay pod securit…

Shubham-astro:
## Description

Release 1.17.31

### DIFF
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -479,3 +479,16 @@ proxy_buffering on;
 {{- end }}
 proxy_set_header            Cookie                  $http_cookie;
 {{- end -}}
+
+{{/*
+git-sync-relay pod-level securityContext
+Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
+*/}}
+{{- define "gitSyncRelay.podSecurityContext" -}}
+{{- if .Values.openshift.enabled }}
+{{- $required := dict "runAsNonRoot" true }}
+{{- merge $required (omit (.Values.gitSyncRelay.securityContext | default dict) "runAsUser" "fsGroup") | toYaml }}
+{{- else }}
+{{- toYaml (.Values.gitSyncRelay.securityContext | default dict) }}
+{{- end -}}
+{{- end }}
--- a/tests/chart/test_git_sync_relay_deployment.py
+++ b/tests/chart/test_git_sync_relay_deployment.py
@@ -304,6 +304,62 @@ class TestGitSyncRelayDeployment:
         assert doc["metadata"]["name"] == "release-name-git-sync-relay"
         assert gsrsecuritycontext == doc["spec"]["template"]["spec"]["securityContext"]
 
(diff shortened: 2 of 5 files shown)


values.yaml
```diff
@@ -655,6 +655,9 @@ airflow:
 # Enable security context constraints required for OpenShift
 sccEnabled: false
 
+openshift:
+  enabled: false
```
rishkarajgi, line 659: We already have `sccEnabled` nearby for OpenShift — is this a replacement for it, or do the two flags serve different purposes? If `openshift.enabled` supersedes `sccEnabled`, might be worth a deprecation note or at least a comment here clarifying the relationship.

templates/_helpers.yaml
```diff
@@ -479,3 +479,16 @@ proxy_buffering on;
...
 proxy_set_header            Cookie                  $http_cookie;
 {{- end -}}
+
+{{/*
+git-sync-relay pod-level securityContext
+Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
+*/}}
+{{- define "gitSyncRelay.podSecurityContext" -}}
+{{- if .Values.openshift.enabled }}
+{{- $required := dict "runAsNonRoot" true }}
```
rishkarajgi, line 489: Nit: the `merge` here forces `runAsNonRoot: true` regardless of what the user set in `.Values.gitSyncRelay.securityContext`. That's the right behavior for OpenShift compliance, but there's no test covering the case where a user explicitly sets `runAsNonRoot: false` — worth adding to make the override intent explicit and prevent future regressions.

tests/chart/test_git_sync_relay_deployment.py
```diff
@@ -304,6 +304,62 @@ def test_gsr_deployment_with_securitycontext_overrides(self, kube_version):
         assert doc["metadata"]["name"] == "release-name-git-sync-relay"
         assert gsrsecuritycontext == doc["spec"]["template"]["spec"]["securityContext"]
 
+    def test_gsr_deployment_openshift_strips_incompatible_security_context(self, kube_version):
+        """Test that fsGroup and runAsUser are stripped from pod securityContext when OpenShift is enabled,
```
rishkarajgi, line 308: Nit: might be worth adding a variant of this test where `securityContext` is not set at all (no key in the values dict) — the `| default dict` in the template should handle it, but a test would confirm the nil-security-context path works end-to-end on OpenShift.

rishkarajgi (commented):
Strips `fsGroup` and `runAsUser` from the git-sync-relay pod security context when OpenShift is enabled, delegating the logic to a new `gitSyncRelay.podSecurityContext` helper and introducing an `openshift.enabled` values flag. Two new tests cover both the stripping and the non-OpenShift passthrough paths.

**Verdict:** 💬 A few questions — logic and tests look solid, just a couple of nits and a question about the relationship between the new `openshift.enabled` flag and the existing `sccEnabled`.

PR #732, opened 2026-05-26
Avoid hostname binary in Celery worker probe

bmanan7:
## Description

The default Celery worker liveness probe currently builds the Celery node name with the OS-level `hostname` executable:

```sh
celery inspect ping -d celery@$(hostname)
```

Some minimal container images, including UBI-based Runtime images, do not include that executable. That makes the probe fail before it can check whether the Celery worker is alive.

This wrapper-chart override keeps the same behavior of targeting the worker container's hostname, but reads it with Python's standard `socket.gethostname()` instead:

```sh
celery inspect ping -d celery@$(python -c 'import socket; print(socket.gethostname())')
```

The probe already depends on Python to run Celery, so this avoids adding another OS binary requirement. Kubernetes also documents that the container hostname is available through the `gethostname` function call: https://kubernetes.io/docs/concepts/containers/container-environment/

Because `airflow-chart` consumes the Airflow worker template from the packaged `astronomer/apc-airflow` subchart, this PR applies the mitigation as an APC wrapper-chart default. The permanent template fix is tracked separately in `astronomer/apc-airflow`.

The override also selects the Celery app import path at runtime from the installed `apache-airflow` package metadata, so it remains compatible with older Airflow 2 images and Airflow 3 images without importing Airflow during shell command substitution.

Related:
- apache/airflow#67471
- astronomer/apc-airflow#18

## Validation

- `uv run pytest tests/chart/test_airflow.py -q`
- `prek run --files values.yaml tests/chart/test_airflow.py`
- Rendered the Celery worker deployment with `helm template` and confirmed the generated worker liveness probe command uses `socket.gethostname()`.
- Ran the command substitutions in `quay.io/astronomer/ap-airflow:2.4.3` and confirmed they expand to the expected Celery app path and container hostname.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -134,7 +134,7 @@ airflow:
         - -c
         - >-
           CONNECTION_CHECK_MAX_COUNT=0 exec /entrypoint python -m celery --app
-          $(python -c 'import airflow, re; version = getattr(airflow, "__version__", "0.0"); major, minor = map(int, re.match(r"(\d+)\.(\d+)", version).groups()); print("airflow.providers.celery.executors.celery_executor.app" if (major, minor) >= (2, 7) else "airflow.executors.celery_executor.app")')
+          $(python -c 'import re; from importlib.metadata import version; airflow_version = version("apache-airflow"); major, minor = map(int, re.match(r"(\d+)\.(\d+)", airflow_version).groups()); print("airflow.providers.celery.executors.celery_executor.app" if (major, minor) >= (2, 7) else "airflow.executors.celery_executor.app")')
           inspect ping -d celery@$(python -c 'import socket; print(socket.gethostname())')
     persistence:
       # Enable persistent volumes
--- a/values.yaml
+++ b/values.yaml
@@ -127,6 +127,15 @@ airflow:
 
   # Airflow worker settings
   workers:
+    livenessProbe:
+      # UBI-based Runtime images may not include the hostname binary.
+      command:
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -127,6 +127,15 @@ airflow:
...
   workers:
+    livenessProbe:
+      # UBI-based Runtime images may not include the hostname binary.
+      command:
+        - sh
+        - -c
+        - >-
+          CONNECTION_CHECK_MAX_COUNT=0 exec /entrypoint python -m celery --app
+          $(python -c 'import re; from importlib.metadata import version; airflow_version = version("apache-airflow"); major, minor = map(int, re.match(r"(\d+)\.(\d+)", airflow_version).groups()); print("airflow.providers.celery.executors.celery_executor.app" if (major, minor) >= (2, 7) else "airflow.executors.celery_executor.app")')
+          inspect ping -d celery@$(python -c 'import socket; print(socket.gethostname())')
```
pgvishnuram, line 138: since this is compactible with both ubi9 and regular images it can go directly in apc-airflow

bmanan7: Thanks @pgvishnuram . Closing this PR as the permenant fix lies in PR https://github.com/astronomer/apc-airflow/pull/18.

bmanan7:
Deployed permanent fix in PR https://github.com/astronomer/apc-airflow/pull/18.

PR #735, opened 2026-06-10
add support for Istio ingress

htech7x:
## Description
Adding support for Istio ingress

## Related Issues
https://linear.app/astronomer/issue/APC-1423/add-optional-istio-ingress-support-to-apc-helm-chart

## Testing
Testing was conducted for APC 1.1.5 in unified mode.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -162,7 +162,7 @@ spec:
                   name: flower-ui
                   {{- end }}
 {{- end }}
-{{- if and .Values.dagDeploy.enabled  }}
+{{- if and .Values.dagDeploy.enabled (not .Values.airflow.istio.enabled) }}
 ---
 kind: Ingress
 apiVersion: networking.k8s.io/v1
--- a/templates/webserver/webserver-virtualservice.yaml
+++ b/templates/webserver/webserver-virtualservice.yaml
@@ -46,29 +46,29 @@ spec:
     - {{- include "deployments_subdomain" . | indent 1 }}
   {{- if hasKey .Values.airflow.istio "gateway" }}
   gateways:
-  - {{ .Values.airflow.istio.gateway }}
+    - {{ .Values.airflow.istio.gateway }}
   {{- end }}
   http:
-  # Webserver routes
-  - match:
-    - uri:
-        prefix: "/{{ .Release.Name }}/airflow"
-    route:
-    - destination:
-        host: {{ $fullname }}-webserver.{{ .Release.Namespace }}.svc.cluster.local
-        port:
-        {{- if .Values.authSidecar.enabled  }}
-          number: {{ .Values.authSidecar.port }}
-        {{- else }}
-          number: {{ .Values.airflow.ports.airflowUI }}
-        {{- end }}
-    {{- if hasKey .Values.airflow.istio "virtualService" }}
-    {{- if hasKey .Values.airflow.istio.virtualService "timeout" }}
-    timeout: {{ .Values.airflow.istio.virtualService.timeout }}
-    {{- end }}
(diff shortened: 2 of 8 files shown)


pgvishnuram:
@htech7x git sync relay virtualhost missing we are using that for receiving webhook

pgvishnuram:
added missing gitsync templates for webhook

pgvishnuram:
2 decisions 
pending since istio needs authsidecar as a hard requirement - we should decided to link with authsidecar enabled - this can be injected from houston api itself 

instead of airflow.istio.enabled should we move to istio.enabled making it as astronomer managed

pgvishnuram (approved):

templates/webserver/webserver-virtualservice.yaml
```diff
@@ -0,0 +1,74 @@
...
+  {{- with .Values.airflow.webserver.labels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .Values.airflow.webserver.annotations }}
+  annotations: {{- toYaml .Values.airflow.webserver.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  hosts:
+    - {{- include "deployments_subdomain" . | indent 1 }}
+  {{- if hasKey .Values.airflow.istio "gateway" }}
```
rishkarajgi, line 47: `hasKey` is true even if `airflow.istio.gateway` is set to an empty string, and I confirmed this renders `gateways:\n  - ` (an empty gateway entry) -- Istio will reject that. Same pattern in the other four VirtualService templates. A truthy check (`with`/`if .Values.airflow.istio.gateway`) instead of `hasKey` would skip the block instead of emitting a broken entry.

templates/ingress.yaml
```diff
@@ -256,7 +256,7 @@ spec:
                   {{- end }}
     {{- end }}
 {{- end }}
-{{- if and .Values.dagDeploy.enabled  }}
+{{- if and .Values.dagDeploy.enabled (not .Values.airflow.istio.enabled) }}
```
rishkarajgi, line 259: This guard is only added here and on the git-sync-relay webhook ingress -- the main webserver and flower Ingress blocks earlier in this file have no istio-enabled suppression at all. Confirmed by grep: this is the only `istio` reference in the whole file. With both `airflow.istio.enabled` and the default ingress on, you'd get a Kubernetes Ingress and an Istio VirtualService routing the same host/path, which seems to work against the point of the feature (customers routing through their own mesh, not a mix of both).

rishkarajgi (commented):
Went through Copilot's two rounds of comments here first, since its headline claim (indentation makes `gateways`/`http` invalid YAML, `uri` used as a list of `path`) sounded serious enough to check directly. Rendered all four VirtualService templates with `helm template` and ran `helm lint` on the whole chart with Istio enabled -- everything renders as clean, valid YAML, `uri` is correctly an object with `prefix`/`regex`/`exact`, and lint passes with 0 failures. That claim is wrong, which is probably why it didn't block the approval -- but two smaller, real issues were in the same comment batch and are still open, noted inline.

One thing worth flagging for whoever's coordinating this with the companion PR (astronomer#3290, same Linear ticket APC-1423): that PR's Istio toggle is `global.istioEnabled` (flat, platform-level), this one's is `airflow.istio.enabled` (nested, per-deployment) -- different shape for what a customer configuring "turn on Istio" would reasonably expect to be one setting. Worth confirming that's intentional (different scopes: platform ingress vs. per-airflow-deployment routing) rather than just uncoordinated naming between the two PRs.

Also worth noting, same as Copilot's comment on this: no `tests/chart_tests/*.py` coverage for any of the five new VirtualService templates or the new ingress-suppression conditions, so a real regression in either wouldn't be caught by CI.

PR #736, opened 2026-06-18
PINF-784: support dots in dag_id and task_id for Vector log path parsing in sidecar logging.

Shubham-astro:
## Description

Task logs from TaskGroup tasks were silently dropped by Vector when using the Kubernetes executor with DS logging. This happened as the regex in the parse_airflowV3_path transform used [a-zA-Z0-9_-] for dag_id and task_id path segments. Airflow natively permits dots in these identifiers, and TaskGroup task IDs use dot notation by default (prefix_group_id=True), e.g. my_group.my_task. Vector failed to match these paths and dropped the logs entirely.

## Related Issues

https://linear.app/astronomer/issue/PINF-784/task-logs-from-taskgroup-tasks-with-prefix-group-idtrue-the-default

## Testing

-  Tested the fix on the cluster
Screenshot 2026-06-18 at 12 29 40 PM

- With latest cm:
Screenshot 2026-06-18 at 12 30 26 PM

- Also confirmed tha other tasks were able to emit logs without issues:
Screenshot 2026-06-18 at 12 31 19 PM

## Merging

Master, 1.17

### DIFF
--- a/templates/logging-sidecar-configmap.yaml
+++ b/templates/logging-sidecar-configmap.yaml
@@ -34,7 +34,7 @@ data:
         {{- if semverCompare ">=3.0.0" .Values.airflow.airflowVersion }}
           # Airflow 3.x - Parse JSON structured logs
           parsed_log = parse_json(.message) ?? .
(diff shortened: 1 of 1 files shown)


karankhanchandani (approved):
LGTM

PR #738, opened 2026-06-25
Add llm dev files

danielhoherd:
## Description

- Add LLM files to aid in all development.
- Fix CODEOWNERS

## Related Issues

This work was done as part of <https://linear.app/astronomer/issue/PINF-585> but is more broadly useful.

## Testing

This is all dev workflow, no code changes to the product are included. No testing is needed.

## Merging

This should be cherry-picked to all supported branches so we can better use LLMs in all branches.


adhipjos (approved):
Looks good

PR #740, opened 2026-06-26
Restore AF3 version gate for log path inclusion

Shubham-astro:
## Description

Previously, `/usr/local/airflow/logs/**/*.log` was unconditionally included in the Vector sidecar log sources. This path only exists in Airflow 3.x deployments. and including it for AF2 emits more logs for AF3, causing cost hike for customers specially using external elasticsearch.

## Related Issues

Do not merge this PR until this text is replaced with links to related issues.

## Testing

- Added unittests.
- This is what we had before as well, no QA needed. 

## Merging

Master, 1.17

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 # apiVersion v2 is Helm 3
 apiVersion: v2
 name: airflow
-version: 1.17.35
+version: 1.17.36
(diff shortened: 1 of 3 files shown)


pgvishnuram:
bump the chart version too

pgvishnuram (approved):

PR #743, opened 2026-06-29
PINF-425 add https git-sync from private repo

danielhoherd:
## Description

- **`values.yaml`**: new `gitSyncRelay.repo.auth` block.
  - `auth.type`: one of `ssh`, `https-pat`, `https-none`. Unset implies SSH when `sshPrivateKeySecretName` is set (back-compat), otherwise no auth.
  - `auth.https.credentialsSecretName`: name of the existing Secret holding HTTPS credentials (houston creates the per-deployment `${releaseName}-git-sync` Secret via Commander, same Secret used for SSH).
  - `auth.https.credentialsSecretKey`: keys the relay reads from the mounted credentials dir (`token: gitHttpsToken`, `username: gitHttpsUsername`).

- **`templates/git-sync-relay/git-sync-relay-deployment.yaml`**:
  - Resolves the auth method from `repo.auth.type`, treating an SSH key with no explicit type as SSH.
  - Fails the render if both SSH and HTTPS are configured, since the two are mutually exclusive per deployment.
  - For `https-pat`: adds a `git-https-secret` volume and mounts it at `/etc/git-secret/https`. Mounted as a directory (no `subPath`) so kubelet propagates rotated credentials without a pod restart (FR2.2).
  - Sets `GIT_SYNC_AUTH_TYPE` for both HTTPS modes and `GIT_SYNC_HTTPS_SECRET_DIR` for `https-pat`.

- **`tests/chart/test_git_sync_relay_deployment.py`**: render tests covering the HTTPS paths and the existing SSH/no-auth defaults.
  - `https-pat`: secret volume, mount path, `readOnly`, env vars, and absence of SSH artifacts.
  - `https-pat` mount has no `subPath`.
  - `https-none`: no secret volume, `GIT_SYNC_AUTH_TYPE` set, no secret dir or SSH env.
  - SSH + HTTPS together fails the render.
  - SSH-with-credentials and no-credentials defaults still render as before.

## Related Issues

- <https://linear.app/astronomer/issue/PINF-425/fr11-airflow-chart>

## Testing

Tests have been updated but this has not been manually tested in a cluster.

## Merging

This is only for 2.1

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ auto_update:
   cooldown_days: 7
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.15.9"
+    rev: "v0.15.18"
     hooks:
       - id: ruff-check
         args:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,7 @@
 ---
 exclude: '^(venv|\.vscode|tests/k8s_schema|tests/chart/test_data)' # regex
+auto_update:
+  cooldown_days: 7
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
     rev: "v0.15.9"
--- a/values.yaml
+++ b/values.yaml
@@ -928,11 +928,12 @@ gitSyncRelay:
         # Name of an existing Secret holding the HTTPS credentials. Not maintained
         # by this chart — houston creates the per-deployment ${releaseName}-git-sync
         # Secret via Commander (the same Secret used for SSH).
+        #
+        # The Secret must contain the keys `gitHttpsToken` and `gitHttpsUsername`.
+        # These names are a fixed contract, not configurable: git-sync-relay reads
+        # them as filenames from the mounted directory (PINF-426) and houston writes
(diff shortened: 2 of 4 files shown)


values.yaml
```diff
@@ -917,6 +917,22 @@ gitSyncRelay:
...
+      type: ~
+      https:
+        # Name of an existing Secret holding the HTTPS credentials. Not maintained
+        # by this chart — houston creates the per-deployment ${releaseName}-git-sync
+        # Secret via Commander (the same Secret used for SSH).
+        credentialsSecretName: ~
+        # Secret keys the relay reads from the mounted credentials directory.
+        credentialsSecretKey:
+          token: gitHttpsToken
+          username: gitHttpsUsername
```
Simpcyclassy: Are these read somewhere in the deployment template?

danielhoherd: No. I removed them and added a comment on how they work.

edmund-astronomer (approved):

.pre-commit-config.yaml
```diff
@@ -1,8 +1,10 @@
 ---
 exclude: '^(venv|\.vscode|tests/k8s_schema|tests/chart/test_data)' # regex
+auto_update:
+  cooldown_days: 7
```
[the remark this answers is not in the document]
danielhoherd: This is a prek key

PR #744, opened 2026-07-01
PINF-810: dual-host deployment Ingress for CP HA global-parent alias

adhipjos:
## Description

Paired chart change for [PINF-810](https://linear.app/astronomer/issue/PINF-810/deployment-url-alias-urls-for-existing-customer-ha-migration) (Alias URLs). Houston supplies a per-cluster global parent. The airflow-chart renders the actual deployment `Ingress` host rules and TLS SANs, so the dual-host rendering lives here.

Adds an optional `ingress.globalBaseDomain` value (set by houston from the per-cluster `globalBaseDomain`). When set, each deployment Ingress (webserver, flower, dag-server) emits a **parallel `host:` rule and TLS SAN under the global parent** alongside the existing per-CP `baseDomain` rules. This ensures browser flows reach Airflow on the global host (covered by the `.<globalBaseDomain>` cookie) while old per-CP URLs keep resolving for tooling.

### Changes
- `values.yaml`: new `ingress.globalBaseDomain` (default `~`).
- `templates/_helpers.yaml`: global variants of the `deployments_subdomain` / `airflow_subdomain` / `flower_subdomain` helpers.
- `templates/ingress.yaml`: gated global host rules + TLS hosts on all three Ingress objects. Everything is behind `{{- if .Values.ingress.globalBaseDomain }}`, so **non-HA output (no `globalBaseDomain`) is the same**.
- `tests/chart/test_ingress.py`: dual-host rendering when set, single-host (unchanged) when unset.

## Related Issues

[Related astronomer/issues#PINF-810](https://linear.app/astronomer/issue/PINF-810/deployment-url-alias-urls-for-existing-customer-ha-migration). Pairs with the houston-api change (astronomer/houston-api#2725) that emits `ingress.globalBaseDomain`.

## Testing

- `make unittest-chart` / `pytest tests/chart/test_ingress.py`: **50 passing** (existing + 2 new global-alias tests × 5 k8s versions).
- `helm template` verified: dual hosts + dual TLS SANs on all three Ingress objects when `ingress.globalBaseDomain` is set; zero global hosts when unset.

## Merging

Target: `feature/cp-ha-dr`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,7 +172,7 @@ jobs:
   run_pre_commit:
     resource_class: small
     docker:
-      - image: quay.io/astronomer/ci-pre-commit:2026-04
+      - image: quay.io/astronomer/ci-pre-commit:2026-07
     steps:
       - checkout
       - run:
@@ -207,7 +207,7 @@ jobs:
 
   unittest-charts:
     docker:
-      - image: quay.io/astronomer/ci-helm-release:2026-04
+      - image: quay.io/astronomer/ci-helm-release:2026-07
     parallelism: 8
     steps:
       - setup_remote_docker:
@@ -226,7 +226,7 @@ jobs:
 
   build-and-release-internal:
     docker:
-      - image: quay.io/astronomer/ci-helm-release:2026-04
+      - image: quay.io/astronomer/ci-helm-release:2026-07
     steps:
       - checkout
       - run:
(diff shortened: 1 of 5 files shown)


danielhoherd (approved):
Looks good

PR #751, opened 2026-07-14
PINF-774 git-sync private-ca trust

danielhoherd:
## Description

**PINF-774: Wire private-CA trust into the git-sync-relay deployment**

Mount customer private CAs into the git-sync container and run `update-ca-certificates`, so HTTPS+PAT clone works against private-CA-signed Git hosts. All wiring is gated on `global.privateCaCerts` (empty by default, so existing renders are unchanged). Mirrors the platform components' pattern.

## Changes

- `templates/git-sync-relay/git-sync-relay-deployment.yaml` (gated on `global.privateCaCerts`):
  - writable `etc-ssl-certs` emptyDir mounted over `/etc/ssl/certs`
  - `etc-ssl-certs-copier` initContainer to seed it from the image's certs
  - mount each CA Secret under `/usr/local/share/ca-certificates/<name>.pem`
  - set `UPDATE_CA_CERTS=true` on the git-sync container
- `templates/_helpers.yaml`: new `gitSyncRelay.customCaVolumes` / `gitSyncRelay.customCaVolumeMounts` helpers ranging over `global.privateCaCerts` (each Secret keyed `cert.pem`).
- `values.yaml`: document `global.privateCaCerts`.

## Notes

- Non-root / OpenShift safe: `/etc/ssl/certs` is a writable emptyDir made group-writable by the pod `fsGroup`, so `update-ca-certificates` runs without root.

## Related Issues

- https://linear.app/astronomer/issue/PINF-774

## Testing

I have manually tested https private-ca trust end-to-end in k3d against a local forgejo instance.

## Merging

This is only for APC 2.1 (airflow-chart version 1.18)

### DIFF
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -512,26 +512,68 @@ Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
 {{/*
 Private-CA trust for the git-sync container. When global.privateCaCerts is set
 (each entry is a Secret with a `cert.pem` key, synced into this deployment's
-namespace by commander), mount each bundle under /usr/local/share/ca-certificates
-so update-ca-certificates trusts it. Mirrors the astronomer chart's
+namespace by commander), the git-sync container trusts those CAs: each bundle is
+mounted under /usr/local/share/ca-certificates and update-ca-certificates runs
+over a writable emptyDir overlay of /etc/ssl/certs, so it works as the non-root
+relay user (readOnlyRootFilesystem is enforced). Mirrors the astronomer chart's
 custom_ca_volumes / custom_ca_volume_mounts so the same operator-provided CA
-Secrets work here. Paired with a writable emptyDir over /etc/ssl/certs (see the
-deployment template) so update-ca-certificates can run as the non-root relay user.
+Secrets work here.
+
+The four helpers below emit the volumes, the copier initContainer, the volume
+mounts, and the UPDATE_CA_CERTS env. Each is a no-op unless global.privateCaCerts
+is set, so callers include them unconditionally (no repeated guard at the call
+site).
 */}}
 {{- define "gitSyncRelay.customCaVolumes" }}
+{{- if .Values.global.privateCaCerts }}
+- name: etc-ssl-certs
+  emptyDir: {}
 {{- range .Values.global.privateCaCerts }}
 - name: {{ . }}
   secret:
     secretName: {{ . }}
 {{- end }}
 {{- end }}
+{{- end }}
+
+{{- define "gitSyncRelay.customCaInitContainer" }}
+{{- if .Values.global.privateCaCerts }}
+- name: etc-ssl-certs-copier
+  image: "{{ .Values.gitSyncRelay.images.gitSync.repository }}:{{ .Values.gitSyncRelay.images.gitSync.tag }}"
+  securityContext: {{- toYaml .Values.gitSyncRelay.gitSync.securityContext | nindent 4 }}
+  command:
+    - sh
+    - -c
+    - |
+      if [ -d /etc/ssl/certs ]; then
+        cp -r /etc/ssl/certs/* /etc/ssl/certs_copy/
+      else
+        echo "No /etc/ssl/certs directory found, skipping copy."
+      fi
+  volumeMounts:
+    - name: etc-ssl-certs
+      mountPath: /etc/ssl/certs_copy
+  resources: {{- toYaml .Values.gitSyncRelay.gitSync.resources | nindent 4 }}
+{{- end }}
+{{- end }}
 
 {{- define "gitSyncRelay.customCaVolumeMounts" }}
+{{- if .Values.global.privateCaCerts }}
+- name: etc-ssl-certs
+  mountPath: /etc/ssl/certs
 {{- range $secret_name := .Values.global.privateCaCerts }}
 - name: {{ $secret_name }}
   mountPath: /usr/local/share/ca-certificates/{{ $secret_name }}.pem
   subPath: cert.pem
   readOnly: true
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{- define "gitSyncRelay.customCaEnv" }}
+{{- if .Values.global.privateCaCerts }}
+- name: UPDATE_CA_CERTS
+  value: "true"
 {{- end }}
 {{- end }}
 
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -570,7 +570,7 @@ The same index is used in customCaVolumeMounts so each mount matches its volume.
   mountPath: /etc/ssl/certs
 {{- range $index, $secret_name := .Values.global.privateCaCerts }}
 - name: private-ca-{{ $index }}
-  mountPath: /usr/local/share/ca-certificates/private-ca-{{ $index }}.crt
+  mountPath: /usr/local/share/ca-certificates/private-ca-{{ $index }}.pem
   subPath: cert.pem
   readOnly: true
 {{- end }}
--- a/templates/_helpers.yaml
+++ b/templates/_helpers.yaml
@@ -524,14 +524,21 @@ mounts, and the UPDATE_CA_CERTS env. Each is a no-op unless global.privateCaCert
 is set, so callers include them unconditionally (no repeated guard at the call
 site).
 */}}
+{{/*
+Volume names must be DNS-1123 labels (no dots, <=63 chars, unique), but Secret
+names are DNS-1123 subdomains (may contain dots, up to 253 chars). Derive the
+volume name from the list index (private-ca-N) so it is always valid and can
+never collide; the real Secret name stays on `secretName` / the mount filename.
+The same index is used in customCaVolumeMounts so each mount matches its volume.
+*/}}
 {{- define "gitSyncRelay.customCaVolumes" }}
 {{- if .Values.global.privateCaCerts }}
 - name: etc-ssl-certs
   emptyDir: {}
-{{- range .Values.global.privateCaCerts }}
-- name: {{ . }}
+{{- range $index, $secret_name := .Values.global.privateCaCerts }}
+- name: private-ca-{{ $index }}
   secret:
-    secretName: {{ . }}
+    secretName: {{ $secret_name }}
 {{- end }}
 {{- end }}
 {{- end }}
@@ -546,7 +553,7 @@ site).
     - -c
     - |
       if [ -d /etc/ssl/certs ]; then
-        cp -r /etc/ssl/certs/* /etc/ssl/certs_copy/
+        cp -r /etc/ssl/certs/. /etc/ssl/certs_copy/
       else
         echo "No /etc/ssl/certs directory found, skipping copy."
       fi
@@ -561,9 +568,9 @@ site).
 {{- if .Values.global.privateCaCerts }}
 - name: etc-ssl-certs
   mountPath: /etc/ssl/certs
-{{- range $secret_name := .Values.global.privateCaCerts }}
-- name: {{ $secret_name }}
-  mountPath: /usr/local/share/ca-certificates/{{ $secret_name }}.pem
+{{- range $index, $secret_name := .Values.global.privateCaCerts }}
+- name: private-ca-{{ $index }}
+  mountPath: /usr/local/share/ca-certificates/private-ca-{{ $index }}.crt
   subPath: cert.pem
   readOnly: true
 {{- end }}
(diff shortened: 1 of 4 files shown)


templates/_helpers.yaml
```diff
@@ -509,6 +509,32 @@ Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
...
+- name: {{ . }}
+  secret:
+    secretName: {{ . }}
+{{- end }}
+{{- end }}
+
+{{- define "gitSyncRelay.customCaVolumeMounts" }}
+{{- range $secret_name := .Values.global.privateCaCerts }}
+- name: {{ $secret_name }}
+  mountPath: /usr/local/share/ca-certificates/{{ $secret_name }}.pem
```
edmund-astronomer: does this need to be crt since we are using `update-ca-certificates` in git-sync-relay which i believe does not accept pem
https://github.com/astronomer/git-sync-relay/pull/59/changes

values.yaml
```diff
@@ -2,6 +2,15 @@
...
 # Declare variables to be passed into your templates.
 
+global:
+  # Names of Secrets (each with a `cert.pem` key) holding customer private CA
+  # bundles to trust. Houston/commander populate this per deployment with the
+  # names of the CA Secrets synced into the deployment namespace (via the
+  # astronomer.io/commander-sync annotation). When set, the git-sync container
+  # trusts them via update-ca-certificates (see the git-sync-relay deployment).
+  # Mirrors the astronomer chart's global.privateCaCerts.
+  privateCaCerts: []
```
pgvishnuram, line 12: why a new section here @danielhoherd

templates/git-sync-relay/git-sync-relay-deployment.yaml
```diff
@@ -159,7 +185,16 @@ spec:
             {{- end }}
             - name: tmp
               mountPath: /tmp
+            {{- if .Values.global.privateCaCerts }}
+            - name: etc-ssl-certs
+              mountPath: /etc/ssl/certs
+            {{- include "gitSyncRelay.customCaVolumeMounts" . | nindent 12 }}
+            {{- end }}
```
pgvishnuram: are you sending this volume via houston or using configsyncer to sync the secretnames here

templates/git-sync-relay/git-sync-relay-deployment.yaml
```diff
@@ -159,7 +185,16 @@ spec:
             {{- end }}
             - name: tmp
               mountPath: /tmp
+            {{- if .Values.global.privateCaCerts }}
```
pgvishnuram: call in one place reference to other places

templates/git-sync-relay/git-sync-relay-deployment.yaml
```diff
@@ -159,7 +185,16 @@ spec:
...
+            - name: etc-ssl-certs
+              mountPath: /etc/ssl/certs
+            {{- include "gitSyncRelay.customCaVolumeMounts" . | nindent 12 }}
+            {{- end }}
           env:
+            {{- if .Values.global.privateCaCerts }}
```
pgvishnuram: same here

templates/_helpers.yaml
pgvishnuram: @edmund-astronomer the logic we do here is we mount in the /usr/local and update ca certs creates a copy inside /etc/ssl/certs 

@danielhoherd how are you moving the certs - using configsyncer ??

danielhoherd: @pgvishnuram yes, it uses the built-in config-syncer feature. This is called out in the houston PR.

values.yaml
danielhoherd: This is needed because no airflow components have any privateCaCerts wired in yet. The only pods that trust private-CA are houston injected. So this feature is well known in astronomer, but surprisingly it does not yet exist in airflow-chart.

templates/git-sync-relay/git-sync-relay-deployment.yaml
danielhoherd: The workflow is that the APC cluster operator needs to create the secret in the astronomer namespace, and commander will automatically sync it to the airflow namespaces.

templates/_helpers.yaml
pgvishnuram: customer has to manually add the label when including the privateCA - that can be addressed via documentation

pgvishnuram: there is one cavet here - customer has to call the update cluster config if they change the certs list or change the secret name

danielhoherd: the private-ca secret lives entirely in the DP, nothing in the CP. does that still require updating the cluster config? seems like we could just update the secret and that would be sufficient.

templates/_helpers.yaml
```diff
@@ -509,6 +509,74 @@ Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
...
+*/}}
+{{- define "gitSyncRelay.customCaVolumes" }}
+{{- if .Values.global.privateCaCerts }}
+- name: etc-ssl-certs
+  emptyDir: {}
+{{- range .Values.global.privateCaCerts }}
+- name: {{ . }}
+  secret:
+    secretName: {{ . }}
+{{- end }}
```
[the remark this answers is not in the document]
danielhoherd: Fixed in 9c57162c. Volume names are now derived from the list index (`private-ca-N`), always a valid DNS-1123 label and collision-free; the real Secret name stays on `secretName` and the mount filename. The same index is used for the volume and its mount so they match.

templates/_helpers.yaml
```diff
@@ -509,6 +509,74 @@ Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
...
+
+{{- define "gitSyncRelay.customCaVolumeMounts" }}
+{{- if .Values.global.privateCaCerts }}
+- name: etc-ssl-certs
+  mountPath: /etc/ssl/certs
+{{- range $secret_name := .Values.global.privateCaCerts }}
+- name: {{ $secret_name }}
+  mountPath: /usr/local/share/ca-certificates/{{ $secret_name }}.pem
+  subPath: cert.pem
+  readOnly: true
```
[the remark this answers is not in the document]
danielhoherd: Fixed in 9c57162c — mounted as `private-ca-N.crt` now (content stays PEM via `subPath: cert.pem`), so `update-ca-certificates` picks it up reliably.

templates/_helpers.yaml
```diff
@@ -509,6 +509,74 @@ Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
...
+{{- if .Values.global.privateCaCerts }}
+- name: etc-ssl-certs-copier
+  image: "{{ .Values.gitSyncRelay.images.gitSync.repository }}:{{ .Values.gitSyncRelay.images.gitSync.tag }}"
+  securityContext: {{- toYaml .Values.gitSyncRelay.gitSync.securityContext | nindent 4 }}
+  command:
+    - sh
+    - -c
+    - |
+      if [ -d /etc/ssl/certs ]; then
+        cp -r /etc/ssl/certs/* /etc/ssl/certs_copy/
```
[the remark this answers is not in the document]
danielhoherd: Fixed in 9c57162c: the copier now runs `cp -r /etc/ssl/certs/. /etc/ssl/certs_copy/`, which handles an empty dir and dotfiles.

tests/chart/test_git_sync_relay_deployment.py
```diff
@@ -55,6 +55,49 @@ def test_gsr_deployment_gsr_enabled_with_defaults(self, kube_version):
...
+        assert "etc-ssl-certs-copier" in inits
+        copier_mounts = {m["name"]: m for m in inits["etc-ssl-certs-copier"]["volumeMounts"]}
+        assert copier_mounts["etc-ssl-certs"]["mountPath"] == "/etc/ssl/certs_copy"
+
+        # git-sync container: writable trust store + CA mount + UPDATE_CA_CERTS.
+        c_by_name = get_containers_by_name(docs[0])
+        gs_mounts = {m["name"]: m for m in c_by_name["git-sync"]["volumeMounts"]}
+        assert gs_mounts["etc-ssl-certs"]["mountPath"] == "/etc/ssl/certs"
+        assert gs_mounts["my-private-ca"]["mountPath"] == "/usr/local/share/ca-certificates/my-private-ca.pem"
+        assert gs_mounts["my-private-ca"]["subPath"] == "cert.pem"
```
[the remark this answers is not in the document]
danielhoherd: Updated in 9c57162c to assert the `.crt` mount path.

tests/chart/test_git_sync_relay_deployment.py
```diff
@@ -55,6 +55,49 @@ def test_gsr_deployment_gsr_enabled_with_defaults(self, kube_version):
...
+            values=values,
+        )
+        assert len(docs) == 1
+        spec = docs[0]["spec"]["template"]["spec"]
+
+        # Volumes: writable trust-store overlay + the CA secret.
+        vols = {v["name"]: v for v in spec["volumes"]}
+        assert "etc-ssl-certs" in vols and "emptyDir" in vols["etc-ssl-certs"]
+        assert vols["my-private-ca"]["secret"]["secretName"] == "my-private-ca"
+
```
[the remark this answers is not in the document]
danielhoherd: Updated in 9c57162c to assert the index-derived volume name (`private-ca-0`), plus a new multi-CA test with dotted Secret names so raw names are never used as volume names.

templates/_helpers.yaml
danielhoherd: Switched the mount to `.crt` in 9c57162c so `update-ca-certificates` picks it up reliably (it processes `.crt` under `/usr/local/share/ca-certificates`; content is still PEM). On the cluster-config caveat: the Secret content lives entirely in the DP (commander syncs it into the airflow namespaces), so rotating the cert under the same Secret name needs no cluster-config change. But the list of Secret names is houston cluster-scoped config (`deployments.privateCaCertSecretNames`, emitted as `global.privateCaCerts`), so adding/removing a CA or renaming the Secret does require an `updateCluster`. So @pgvishnuram is right for the name/list case; content rotation is the no-op case.

templates/git-sync-relay/git-sync-relay-deployment.yaml
danielhoherd: Done in 3d934118 — the private-CA wiring is now in named templates (`gitSyncRelay.customCaVolumes` / `customCaInitContainer` / `customCaVolumeMounts` / `customCaEnv`) in `_helpers.yaml`, and the deployment just includes them, so the `global.privateCaCerts` guard is defined once instead of repeated at each call site.

templates/git-sync-relay/git-sync-relay-deployment.yaml
danielhoherd: Same — covered by the named-template refactor in 3d934118 (this block is now the `gitSyncRelay.customCaEnv` template).

templates/_helpers.yaml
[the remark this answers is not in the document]
danielhoherd: Correction to my earlier reply: keeping `.pem` after all (9bc76b5e). update-ca-certificates on our ap-base (Alpine) image picks up the `.pem` file (verified by the private-CA e2e test), and the astronomer chart already mounts these as `.pem`, so we match that for clarity and consistency. `.crt` is the Debian-documented extension for inclusion but is not required on this base image. The index-based volume-name fix stays.

tests/chart/test_git_sync_relay_deployment.py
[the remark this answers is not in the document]
danielhoherd: Correction: reverted the assertion to `.pem` in 9bc76b5e — keeping the `.pem` filename (see the helper thread for why).

templates/_helpers.yaml
danielhoherd: Following up: after checking the man page you linked (which does list `.crt` as the extension for inclusion), we are keeping `.pem` in 9bc76b5e. On ap-base (Alpine) update-ca-certificates picks up the `.pem` file anyway (the private-CA e2e clone succeeds), and the astronomer chart already uses `.pem`, so a PEM file named `.pem` is the clearer and consistent choice. Volume names are still the index-based `private-ca-N` fix.

values.yaml
pgvishnuram: i thought we can reuse astronomer section which was sitting idle without much use case 

```
astronomer:
  images:
    certgenerator:
      repository: quay.io/astronomer/ap-certgenerator
      tag: 0.1.15
```

pgvishnuram: not a blocker moving forward we have to use one common  core platform label that should be used for all the control plane fused configs today it sits in multiple places

pgvishnuram (approved):

PR #758, opened 2026-07-24
PINF-1085 - adopt airflow ingress global flag

pgvishnuram:
## Description

adopt airflow ingress global flag to controll all ingress object provisioning.
current we only use them for airflow components . This PR extends them for gitsync and dagserver components to have one single flag to disable all

## Related Issues

https://linear.app/astronomer/issue/PINF-1085/adopt-airflow-ingress-global-flag-across-gitsync-dagserver-and-dag

## Testing

when on helm.ingress.enabled is set to false it disables ingress provisioning across all the provisioned components

## Merging

merge to master and release-1.18 , release-1.17


danielhoherd (approved):
Looks good. Mostly this is just moving code around, so the functional diff isn't as big as it seems.

PR #764, opened 2026-08-18
PINF-816: default readinessProbe on git-sync-relay's git-sync container

danielhoherd:
## Description

[PINF-816](https://linear.app/astronomer/issue/PINF-816): companion chart-side change to git-sync-relay's readiness-tracking fix (astronomer/git-sync-relay#76): that PR makes `/readyz` reflect the relay's most recent sync attempt instead of only its first one at startup, but nothing wired that endpoint to an actual k8s probe — the `gitSyncRelay.gitSync` container had no `readinessProbe`/`livenessProbe` default at all (both `{}` in `values.yaml`).

Without a probe, a Deployment's Gitsync URL edited to something invalid produced no k8s-visible signal: the relay pod stayed Ready forever once its first sync had succeeded, silently failing every later poll/webhook cycle in the background.

**Fix:** a hardcoded `readinessProbe` fallback (`httpGet /readyz` on `gitSync.webhookPort`), following the same `{{- if <values> }}...{{- else }}<default>{{- end }}` pattern this template already uses for the `git-daemon` container's own liveness/startup defaults. Still fully overridable via `gitSyncRelay.gitSync.readinessProbe`.

Deliberately **readinessProbe only, not livenessProbe**: restarting the pod cannot fix a bad repo URL or revoked credentials, so tying liveness to sync health would just crash-loop the pod pointlessly instead of surfacing the actual problem.

This closes the loop with commander's **existing** `WaitForDeployment`/`WaitForReady` mechanism with **no commander-side changes needed**: it already discovers Deployments by `release=<name>,tier=airflow` (this Deployment already carries that label) and blocks/times out (default 300s) on `status.readyReplicas < desired`, which depends on pod readinessProbe passing. So a bad URL now makes this readinessProbe fail → `readyReplicas` drops → `WaitForDeployment` times out → commander reports a real, named deploy failure instead of a silently-degraded pod.

Kubelet probe traffic bypasses this chart's restrictive `git-sync-relay` NetworkPolicy (probes go from node to pod directly, never through the CNI's policy-enforced path) — confirmed this is standard behavior, not something needing an explicit allow-rule here.

## Related Issues

- [PINF-816](https://linear.app/astronomer/issue/PINF-816)
- Companion PR: astronomer/git-sync-relay#76 (the actual `/readyz` logic this probe depends on)

## Testing

- `tests/chart/test_git_sync_relay_deployment.py`'s default-values test updated to assert the new `readinessProbe` body instead of asserting its absence; the existing "no hardcoded fallback, customer-configurable only" comment/assertions for `livenessProbe`/`startupProbe` are unchanged (still true — only `readinessProbe` gets a default). The explicit-override test elsewhere in the same file (which sets `readinessProbe`/`livenessProbe` directly) is unaffected since it never hits the new `{{- else }}` branch.
- `uv run pytest tests/chart/test_git_sync_relay_deployment.py` — 155 passed.
- `uv run pytest tests/chart/` (full suite) — 1090 passed, 46 skipped (pre-existing).

### DIFF
--- a/templates/git-sync-relay/git-sync-relay-deployment.yaml
+++ b/templates/git-sync-relay/git-sync-relay-deployment.yaml
@@ -220,6 +220,19 @@ spec:
           {{- end }}
           {{- if .Values.gitSyncRelay.gitSync.readinessProbe }}
           readinessProbe: {{- tpl (toYaml .Values.gitSyncRelay.gitSync.readinessProbe) . | nindent 12 }}
+          {{- else }}
+          # Reflects the MOST RECENT sync attempt (initial, poll, or webhook), not just the
+          # pod's first one at startup -- so a Deployment's Gitsync URL edited to something
+          # invalid after the relay is already running marks this pod NotReady (PINF-816)
+          # instead of staying silently "ready" forever. Deliberately not a livenessProbe: a
+          # restart cannot fix a bad URL and would just crash-loop the pod pointlessly.
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: {{ .Values.gitSyncRelay.gitSync.webhookPort }}
+            initialDelaySeconds: 5
+            periodSeconds: 15
+            failureThreshold: 3
           {{- end }}
           {{- if .Values.gitSyncRelay.gitSync.startupProbe }}
           startupProbe: {{- tpl (toYaml .Values.gitSyncRelay.gitSync.startupProbe) . | nindent 12 }}
--- a/templates/git-sync-relay/git-sync-relay-deployment.yaml
+++ b/templates/git-sync-relay/git-sync-relay-deployment.yaml
@@ -226,13 +226,21 @@ spec:
           # invalid after the relay is already running marks this pod NotReady (PINF-816)
           # instead of staying silently "ready" forever. Deliberately not a livenessProbe: a
           # restart cannot fix a bad URL and would just crash-loop the pod pointlessly.
+          # failureThreshold: 1 is deliberate, not an oversight: /readyz already reflects a
+          # single failed sync cycle, and that sync-level signal is already debounced by
+          # sync_logic()'s own internal fetch retries/backoff before it ever flips -- stacking
+          # a second, multi-tick smoothing layer here (the k8s default is 3) would just delay
+          # surfacing an already-confirmed failure by up to 2 more periodSeconds for no real
+          # gain, since a bare probe-request failure (as opposed to a real 503) should be rare:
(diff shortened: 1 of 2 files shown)


templates/git-sync-relay/git-sync-relay-deployment.yaml
```diff
@@ -220,6 +220,27 @@ spec:
...
+          # instead of staying silently "ready" forever. Deliberately not a livenessProbe: a
+          # restart cannot fix a bad URL and would just crash-loop the pod pointlessly.
+          # failureThreshold: 1 is deliberate, not an oversight: /readyz already reflects a
+          # single failed sync cycle, and that sync-level signal is already debounced by
+          # sync_logic()'s own internal fetch retries/backoff before it ever flips -- stacking
+          # a second, multi-tick smoothing layer here (the k8s default is 3) would just delay
+          # surfacing an already-confirmed failure by up to 2 more periodSeconds for no real
+          # gain, since a bare probe-request failure (as opposed to a real 503) should be rare:
+          # /readyz is a lightweight in-memory read that doesn't block on the sync itself,
+          # which runs in a separate executor thread.
```
pgvishnuram: no body will read this - shorten these ones

danielhoherd: Already trimmed this down (13 lines to 5) in 8cc1e032, before seeing this comment -- same call. Let me know if it still reads as too much.

templates/git-sync-relay/git-sync-relay-deployment.yaml
```diff
@@ -220,6 +220,19 @@ spec:
...
+          {{- else }}
+          # /readyz reflects the most recent sync, so a bad Gitsync URL edit marks this pod
+          # NotReady instead of staying "ready" forever (PINF-816). Not a livenessProbe -- a
+          # restart can't fix a bad URL. failureThreshold: 1 (not the k8s default of 3): the
+          # sync itself already retries internally before /readyz ever flips, so no extra
+          # smoothing is needed here.
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: {{ .Values.gitSyncRelay.gitSync.webhookPort }}
```
pgvishnuram, line 232: regardless of webhook we have /readyz for both poll and webhook right . If thats correct this PR is good to  go.

pgvishnuram (approved):

templates/git-sync-relay/git-sync-relay-deployment.yaml
danielhoherd: /readyz is available in both modes, so we're good.

PR #765, opened 2026-09-01
PINF-1218 - fix custom config sidecar logging

pgvishnuram:
## Description

This PR fixes the sidecar logging custom config feature to support git sync relay and dag server component using shared helper

## Related Issues

https://linear.app/astronomer/issue/PINF-1218/fix-dag-server-custom-sidecar-config-secret-mismatch

## Testing

- covered with unit tests

## Merging

merge to master and release-1.18


templates/_helpers.yaml
```diff
@@ -621,3 +621,12 @@ Strips fsGroup and runAsUser on OpenShift to comply with restricted-v2 SCC.
...
 {{- end }}
+
+
+{{- define "loggingSidecar.configName" -}}
+{{- if .Values.loggingSidecar.customConfig -}}
+sidecar-config
+{{- else -}}
+{{ .Release.Name }}-sidecar-config
+{{- end -}}
+{{- end }}
```
danielhoherd, line 632: Why do we have this distinction? Also, the configMap name `sidecar-config` is really broad. Which sidecar? We should go with something more specific, like `logging-sidecar-config`

pgvishnuram: this is a existing change - we are fixing exiting behaviour here to address the bug.

pgvishnuram: https://www.astronomer.io/docs/astro-private-cloud/v-2-x/export-task-logs#export-logs-using-container-sidecars

Shubham-astro (approved):
LGTM, This is an existing behaviour, no new changes added.

